PR #350, opened 2014-12-18
Allocation-client subpage

kula1922:
[DO NOT MERGE] only for review


src/ralph_scrooge/rest/allocationclient.py
```diff
@@ -8,253 +8,231 @@
...
-        )
-        _clear_daily_usages(
-            pricing_objects,
-            service_usage_type.usage_type,
-            first_day,
-            last_day,
-        )
-        for row in post_data['rows']:
+            for row in post_data['rows']:
+                service_environments = ServiceEnvironment.objects.filter(
```
mkurek: https://github.com/allegro/ralph_pricing/blob/develop/src/ralph_scrooge/models/service.py#L311 - `ServiceEnvironment` has unique_together on service and environment, so your `filter` will return at most 1 record, so you can use `get` instead of `filter` (then no need to iterate below).

src/ralph_scrooge/rest/allocationclient.py
```diff
@@ -8,253 +8,234 @@
...
-                        type=service_usage_type.usage_type,
+            other_type = ExtraCostType.objects.get(id=1)
+            ids = set()
+            for row in post_data['rows']:
+                ids.add(row.get('id'))
+                extra_cost = ExtraCost.objects.get_or_create(
+                    id=row.get('id'),
+                    start=first_day,
+                    end=last_day,
+                    service_environment=service_environment,
```
mkurek: service_environment should be in defaults - what if record has id, but i'll change service_env in GUI? Probably you'll get exception here.

src/ralph_scrooge/rest/allocationclient.py
```diff
@@ -8,253 +8,231 @@
...
+            pricing_service = PricingService.objects.get(services=service)
+        except PricingService.DoesNotExist:
+            pricing_service = PricingService.objects.create(
+                name=service.name + '_pricing_service',
+                symbol=(
+                    service.symbol or
+                    service.name.lower().replace(' ', '_') + '_pricing_service'
+                )
+            )
+            pricing_service.services.add(service)
```
mkurek: that's right - probably i didn't see `objects.create` last time

PR #357, opened 2015-01-13
Feature/calc pricing service include diff

mkurek:
* improved pricing service plugin selection; renamed ceilometer plugin to fixed price plugin
* added charging pricing service by difference between real and calculated costs (ex. from fixed prices)

### DIFF
--- a/src/ralph_scrooge/models/usage.py
+++ b/src/ralph_scrooge/models/usage.py
@@ -9,7 +9,6 @@ from django.core.exceptions import ValidationError
 from django.db import models as db
 from django.utils.translation import ugettext_lazy as _
 
-
 from ralph_scrooge.models.base import BaseUsage, BaseUsageType
 
 PRICE_DIGITS = 16
--- a/src/ralph_scrooge/models/usage.py
+++ b/src/ralph_scrooge/models/usage.py
@@ -9,6 +9,7 @@ from django.core.exceptions import ValidationError
 from django.db import models as db
 from django.utils.translation import ugettext_lazy as _
 
+
 from ralph_scrooge.models.base import BaseUsage, BaseUsageType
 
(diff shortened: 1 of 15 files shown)


src/ralph_scrooge/models/usage.py
```diff
@@ -9,6 +9,7 @@
 from django.db import models as db
 from django.utils.translation import ugettext_lazy as _
 
+
```
kula1922: Needless line

src/ralph_scrooge/plugins/cost/pricing_service.py
```diff
@@ -504,6 +514,56 @@ def _get_service_dynamic_extra_cost(
...
+                ps_real_cost = plugin_runner.run(
+                    'scrooge_costs',
+                    'pricing_service_plugin',
+                    **args
+                )
+
+                # calculate difference between real cost and
+                diff = ps_real_cost - ps_cost
+                result[ps.id] = diff, {}
+            except (KeyError, AttributeError):
```
kula1922: wrap only code where except conditions could be happen

src/ralph_scrooge/plugins/cost/pricing_service.py
```diff
@@ -504,6 +514,56 @@ def _get_service_dynamic_extra_cost(
...
+        result = {}
+        for ps in pricing_service.charged_by_diffs.all():
+            try:
+                args = dict(
+                    type='total_cost',
+                    date=date,
+                    forecast=forecast,
+                    pricing_service=ps,
+                    collapse=True
+                )
```
kula1922: why not args = {...} ? is there any different?

mkurek: why not dict(...) if there is no difference? :)

PR #361, opened 2015-01-16
added database sync plugin

mkurek:
added database sync plugin to get databases from ralph, which will allow to show them as components in components view and charge them for their usages of resources.

requires: https://github.com/allegro/ralph/pull/1274


kula1922:
:ok:

kula1922:
:ok:

src/ralph_scrooge/models/pricing_object.py
```diff
@@ -424,3 +425,49 @@ class DailyVIPInfo(DailyPricingObject):
...
+                pricing_object=self,
+                database_info=self,
+                date=date,
+                service_environment=self.service_environment,
+            )
+
+
+class DailyDatabaseInfo(DailyPricingObject):
+    parent_device = db.ForeignKey(
+        DailyAssetInfo,
```
quamilek: Why `DailyAssetInfo`?

src/ralph_scrooge/plugins/collect/database.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def get_model(ralph_database):
+    model = PricingObjectModel.objects.get_or_create(
+        model_id=ralph_database['type_id'],
+        type_id=PRICING_OBJECT_TYPES.DATABASE,
+        defaults=dict(
+            name=ralph_database['type'],
+        )
+    )[0]
```
quamilek:
```
model, created = PricingObjectModel.objects.get_or_create(
    model_id=ralph_database['type_id'],
    type_id=PRICING_OBJECT_TYPES.DATABASE,
    defaults=dict(
        name=ralph_database['type'],
    )
)
```
looks better

src/ralph_scrooge/plugins/collect/database.py
```diff
@@ -0,0 +1,162 @@
...
+                database_type
+            )
+        except UnknownServiceEnvironmentNotConfiguredError:
+            msg = 'Unknown service environment not configured for {}'.format(
+                database_type
+            )
+            logger.error(msg)
+            return (False, msg)
+
+    for database_type in settings.DATABASE_TYPES:
```
quamilek: I think that you could use one `for` ?

src/ralph_scrooge/plugins/collect/database.py
```diff
@@ -0,0 +1,162 @@
...
+)
+
+from ralph_scrooge.plugins.collect._exceptions import (
+    UnknownServiceEnvironmentNotConfiguredError,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def get_model(ralph_database):
```
quamilek: Name should be more descriptive, is simillar to: 
https://github.com/allegro/ralph_pricing/pull/361/files#diff-1271648e9b72d66d1b1723e023b7938eR28

src/ralph_scrooge/plugins/collect/database.py
kula1922: Its look like case where if there is no unknown_service_environment for any database then do nothing. Thats true?

src/ralph_scrooge/plugins/collect/database.py
kula1922: looks better? why?

src/ralph_scrooge/plugins/collect/database.py
kula1922: Collects plugins are similar and we know about that. We are going to separate some common methods (like get_model) but now there is no time. This method return model so get_model is a correct namespace for me.

src/ralph_scrooge/plugins/collect/database.py
mkurek: Exactly - if unknown service environment is not set for all of the database types, we don't want to run any further processing.

quamilek:
:ok:

src/ralph_scrooge/models/pricing_object.py
```diff
@@ -424,3 +425,49 @@ class DailyVIPInfo(DailyPricingObject):
...
+        verbose_name=_("parent device"),
+        related_name='databases',
+        null=True,
+        blank=True,
+    )
+
+    class Meta:
+        app_label = 'ralph_scrooge'
+
+    def get_daily_pricing_object(self, date):
```
kula1922: docstring

PR #363, opened 2015-01-20
replaced ci_uid by id in costs report

mkurek:
replaced service environment ci_uid by it's id on costs report to easily distinguish each row by something unique


quamilek:
:ok:

andrzej-jankowski:
:ok:

PR #365, opened 2015-01-21
fixed creating ip info in vip plugin

mkurek:
Choice instance (in this case PRICING_OBJECT_TYPES.IP_ADDRESS) has to be assigned to type_id (not type)


xor-xor:
:ok:

zefciu:
:ok:

andrzej-jankowski:
:ok:

PR #366, opened 2015-01-21
fixed team manager security decorator

mkurek:
fixed wrong orm field in _has_permission_to_team decorator

### DIFF
--- a/src/ralph_scrooge/tests/test_security.py
+++ b/src/ralph_scrooge/tests/test_security.py
@@ -10,10 +10,11 @@ from django.test import TestCase, Client
 from ralph.account.models import Perm
 
 from ralph_scrooge.app import Scrooge
-from ralph_scrooge.models import ServiceOwnership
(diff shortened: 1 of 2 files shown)


xor-xor:
:ok:

andrzej-jankowski:
:ok:

src/ralph_scrooge/tests/test_security.py
```diff
@@ -69,6 +79,12 @@ def _get_report(self):
             Scrooge.url_prefix,
         ))
 
+    def _get_team_allocation(self, team):
+        return self.client.get('/{}/rest/allocationclient/{}/2014/10/'.format(
```
xliiv: I think reversed links are better.. .
Since this is the only comment of mine and it's minor thing, I'll 'ok' it.

mkurek: since this PR is already merged, I'll keep this in mind to fix this in one of the next pull requests.

PR #371, opened 2015-01-22
fixed unicode of dynamic extra cost

mkurek:
changed DynamicExtraCost object representation method (__unicode__) - replaced extra_cost_type by dynamic_extra_cost_type, which is valid field of DynamicExtraCost. Added some unit tests to test it as well.


andrzej-jankowski:
:ok:

quamilek:
:ok:

PR #372, opened 2015-01-24
fixed pricing service costs calculation, improved log details and added possibility to save only first depth costs

mkurek:
* fixed pricing service costs calculation
  * removed unnecessary excluded_services passed to dependent services (excluded services are only relevant on costs distribution, not calculatin total costs of pricing serivce)
  * fixed calculation of costs from services, which charge pricing service by their diff with real costs (previously costs dictionary was basicly updated by diff costs, which causes lost of previously calculated values - now if key is already in dictionary, substraction on costs is done)
  * fixed total_cost method for pricing service, when calculating costs of whole service (for all service environments) - when pricing service has other plugin than universal (ex. fixed price plugin), it's service usages percentage division doesn't matter - especially sum could be different than 100, which causes errors when calculating total cost of this pricing service using universal plugin. Now, when total costs is calculated for all service environment, there is no cost-distribution part - only calculating total costs for all service environments.

* unification of costs plugins log messages
* added possibility to save only first depth costs (direct costs, not intermediate), which could significantly improve costs calculation performance, but intermediate costs (ex. for costs-dependency-graph) will not be saved


andrzej-jankowski:
:ok:

quamilek:
:ok:

PR #373, opened 2015-01-25
allocation: do not create new objects on GET request and search ServiceUsageType including date condition

mkurek:
* fix to not create new UsageType, PricingService and ServiceUsageType on GET request - only during save (on POST request)
* check for ServiceUsageType including date condition - allows to define more than one ServiceUsageType for PricingService and look for only one which match current year and month


andrzej-jankowski:
:ok:

quamilek:
:ok:

PR #374, opened 2015-01-28
improved costs calculation by adding additional cache, db indexes and optimizing SQL queries

mkurek:
* added additional cache to pricing service total cost calculation (to not recalculate it every time pricing service is dependency of other pricing service)
* optimized some SQL queries (ex. support, usages)
* added additional DB indexes (on dailyusage and dailypricingobject)


src/ralph_scrooge/plugins/cost/support.py
```diff
@@ -49,23 +49,29 @@ def costs(
...
+            pricing_object__daily_pricing_objects__service_environment__in=(
+                service_environments
+            )
+        ).values_list(
+            'pricing_object__daily_pricing_objects__service_environment_id',
+            'pricing_object_id',
+            'cost',
+            'forecast_cost',
+            'start',
+            'end',
```
ar4s: Maybe use ``values`` instead of ``values_list`` and rename long name by something like this:
``
SupportCost.objects.extra(select={'readable_name': 'pricing_object__daily_pricing_objects__service_environment_id'}).values('readable_name', ...)
``

src/ralph_scrooge/migrations/0019_dailyusage_index.py
```diff
@@ -0,0 +1,466 @@
...
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Adding index on 'DailyUsage', fields ['type', 'date', 'warehouse']
+        db.create_index(u'ralph_scrooge_dailyusage', ['type_id', 'date', 'warehouse_id'])
+
+        db.create_index(u'ralph_scrooge_dailypricingobject', ['date', 'service_environment_id'])
```
ar4s: Please add comment in ``DailyUsage``'s model about ``index_together``.

src/ralph_scrooge/plugins/cost/support.py
mkurek: Done - i've used namedtuple instead.

src/ralph_scrooge/migrations/0019_dailyusage_index.py
mkurek: Done

ar4s:
:ok:

andrzej-jankowski:
:ok:

PR #375, opened 2015-01-30
Update report_table.html

damian1976:
```thead``` tag causes table header in report table to be displayed twice. Please check if ot can be safely removed.

### DIFF
--- a/src/ralph_scrooge/plugins/collect/openstack_mongo_ceilometer.py
+++ b/src/ralph_scrooge/plugins/collect/openstack_mongo_ceilometer.py
@@ -1,0 +1,168 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import datetime
+import logging
+import time
+from itertools import chain
+
+from django.conf import settings
+from ralph_scrooge.utils.common import memoize
+from pymongo import MongoClient
+
+from ralph.util import plugin
+from ralph_scrooge.models import (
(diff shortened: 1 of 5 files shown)


mkurek:
It's not valid solution - removing thead tag does not solve the real problem here.

First of all your problem is that additional header is not hidden by default. To fix this add `$headerFixed.hide();` in table_fixed_header.js in `init` function.

BUT - actually it doesn't fix anything because in new Scrooge there is fixed top bar, what makes this fixed header not working at all. So for now the best way to (temporary) solve this is removing `fixed-header` css class from `table` in report_table.html and/or removing including `scrooge/js/table_fixed_header.js` in base.html template. 

Please follow these instructions, correct your PR and then it'll be merged (as a temporary fix) and we'll think how to solve this problem from scratch (and think if fixed header is really necessary with table scrolling).

mkurek:
Now squash all commits into one and add some extensive commit message about changes.

kula1922:
src/ralph_scrooge/templates/ralph_scrooge/report_table.html
src/ralph_scrooge/templates/ralph_scrooge/base.html

This 2 files are you feature so plx pull current develop, path it and then make new pull request.

damian1976:
You can ignore these changes. They are not important.

vi4m:
@mkurek decide here

mkurek:
@damian1976 if you want it to be merged please make a pull request only with your changes in src/ralph_scrooge/templates/ralph_scrooge/base.html and src/ralph_scrooge/templates/ralph_scrooge/report_table.html

PR #377, opened 2015-02-10
replaced dynamic extra cost no cost error and fixed team division by 0

mkurek:
* replaced dynamic_extra_cost plugin CostNotFoundError by NoPriceCostError, which is one of base cost plugins exception and is handled properly by Collector
* fixed team division by 0 when daily_cost is 0


andrzej-jankowski:
:ok:

ar4s:
:ok:

PR #378, opened 2015-02-11
Added column alias possibility in components view

mkurek:
Added possibility to specify column (field header) alias in components view by passing tuple of 2 strings instead of one string in COMPONENTS_TABLE_SCHEMA type fields.


kula1922:
:ok:

ar4s:
:ok:

PR #379, opened 2015-02-13
Allow to service division only when there is single service usage type

kula1922:
Allow to service division only when there is single service usage type object for choosen pricing service

### DIFF
--- a/src/ralph_scrooge/rest/allocationclient.py
+++ b/src/ralph_scrooge/rest/allocationclient.py
@@ -33,7 +33,7 @@ from ralph_scrooge.models import (
 )
 
 
-class CannotDetermineValidServiceUsageTypeException(APIException):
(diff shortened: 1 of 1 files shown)


src/ralph_scrooge/rest/allocationclient.py
```diff
@@ -193,7 +183,24 @@ def get(self, request, year, month, service, env, *args, **kwargs):
...
+                    "template": "tabservicedivision.html",
+                    "rows": self._get_service_divison(
+                        service,
+                        year,
+                        month,
+                        first_day,
+                    ),
+                }
+            })
+        except CannotDetermineValidServiceUsageTypeException:
```
mkurek: maybe it's good moment to fix this exception name to `CannotDetermineValidServiceUsageTypeError`?

mkurek:
:ok:

xliiv:
:ok:

PR #380, opened 2015-02-13
excluded creating DailyCost instances from transaction management

mkurek:
excluded creating DailyCost instances in collector transaction. From now there are only 2 operations left in transaction scope: deleting previously saved DailyCosts and saving new instances (created before transaction).


src/ralph_scrooge/plugins/cost/collector.py
```diff
@@ -156,17 +156,19 @@ def process(
             forecast,
             plugins,
         )
+        daily_costs = self._create_daily_costs(date, costs, forecast)
```
kula1922: docs is wrong. 

1) Collect cost from all plugins
2) delete previously saved costs
3) save cost in database in tree format

PR #381, opened 2015-02-13
removed limiting choices in ServiceUsageType.usage_type

mkurek:
removed limiting choices to allow to choose one of base usage types (ex. network, collocation) as a method of pricing service costs distribution


kula1922:
:ok:

xliiv:
:ok:

PR #383, opened 2015-02-17
added total usage column to tenants instances report

mkurek:
added total usage column to tenants instances report


ar4s:
:ok:

andrzej-jankowski:
:ok:

PR #384, opened 2015-02-17
excluded creating DailyCost instances from transaction management & limited DailyCost bulk create batch size

mkurek:
excluded creating DailyCost instances in collector transaction. From now there are only 2 operations left in transaction scope: deleting previously saved DailyCosts and saving new instances (created before transaction).

limited DailyCost bulk create batch size (default to 10000, but configurable in settings by DAILY_COST_CREATE_BATCH_SIZE variable) to avoid race-condition for table lock when saving all objects.


ar4s:
:ok:

andrzej-jankowski:
:ok:

PR #386, opened 2015-02-24
improved caching on all cost types calculation

mkurek:
* added cache on all costs calculations - from now costs for single type is only calculated once (for all services and environments), result of it is cached to the end of worker job and if another plugin is interested in costs for particular service-environments, result for all is taken from cache and is filtered by pointed service-environments
* changed order of plugins to first calculate all basic types and at the end calculate pricing services


src/ralph_scrooge/plugins/cost/dynamic_extra_cost.py
```diff
@@ -21,11 +22,11 @@ def total_cost(self, for_all_service_environments=False, *args, **kwargs):
         service_costs = self.costs(*args, **kwargs)
         return self._get_total_costs_from_costs(service_costs)
 
-    def costs(
+    @memoize(skip_first=True)
```
kula1922: why skip first ?

kula1922:
:ok:

zefciu:
:ok:

PR #389, opened 2015-02-26
fixed some costs report bugs

mkurek:
* turn off filtering by service environment when it's not necessary (when report should be calculated for all service environments)
* fixed presenting changes of business line and profit center on reports: show only unique values
* added updating progress in costs report
* changed report cache timeout to 1 hour


kula1922:
:ok:

PR #395, opened 2015-02-27
Feature/ras 310

ar4s:
With range and new calendar.


src/ralph_scrooge/media/scrooge/app/common/controller.js
```diff
@@ -2,14 +2,37 @@
...
+		stats.init();
+		$scope.menuService = menuService;
+		$scope.stats = stats;
+		$scope.getDictLength = function (dict) {
+			if (typeof(dict) == 'object') {
+				return Object.keys(dict).length;
+			}
+		};
+	}])
+	.controller('menuStatsAdapterCtrl', ['$scope', 'stats', function($scope, stats){
```
kula1922: this adapter is specific for calendar and datepicker so move it to app.menu.controller.js

src/ralph_scrooge/media/scrooge/app/common/directive.js
```diff
@@ -2,10 +2,200 @@
...
\ No newline at end of file
+scrooge
+	.directive('tabsdirective', function() {
+	    return {
+	        restrict: 'EACM',
+	        templateUrl: '/static/scrooge/partials/tabs.html',
+	        replace: true,
+	    };
+	})
+	.directive('scCalendar', ['STATIC_URL', function(STATIC_URL) {
```
kula1922: dito

src/ralph_scrooge/media/scrooge/app/common/controller.js
ar4s: OK

kula1922:
squash

src/ralph_scrooge/media/scrooge/app/common/directive.js
ar4s: I don't think so. ``scCalendar`` is a common component - it's separeted from other logic like menu and return only some date value.

PR #408, opened 2015-03-09
added possibility to specify ceilometer usage prefix

mkurek:
added possibility to specify ceilometer usage prefix for single record in OPENSTACK_CEILOMETER (it could be for example used openstack version, when someone will have different unit prices for different flavors in OS versions).


kula1922:
:ok:

PR #410, opened 2015-03-11
Update report_services_usages.py & base_plugin_report.py

damian1976:
Fixes values rounding to given (in admin) number of decimal places. However  the value type is now changed from float to string, it shouldn't complicate anything.


mkurek:
Could you please squash it into one commit?

PR #411, opened 2015-03-13
fixed team division (%) template

mkurek:
previously in team division tab there was wrong template (teamcosts). From now team division and service division templates are merged into single, taballocationclientdivision template.


ar4s:
:ok:

kula1922:
:ok:

PR #417, opened 2015-03-25
Fix: Printing correct decimal places in base_plugin_report.py and report_services_usages.py

damian1976:
Printed values in base_plugin_report.py and report_services_usages.py are properly formatted using correct number of decimal places set in admin panel


kula1922:
:ok:

PR #427, opened 2015-04-16
Added mising image

ar4s:
This change is require if we using built-in staticfiles versioning in Django
because ``collectstatic`` scan all depends files (e.g. css, img) and is
failure if some file doesn't exist.


xliiv:
:ok:

szok:
:ok:

PR #432, opened 2015-04-20
Updated static files to cached backend.

xliiv:
Recently Ralph-core change static files backend to cached one.
This change make use of cached static files by change url generating
method.
In addition test are fixed by setting non-cache backend.

@ar4s @mkurek


mkurek:
:ok:

mkurek:
@xliiv please take a loot at travis errors

ar4s:
:ok:

PR #435, opened 2015-04-23
Fixed demo data generation.

xliiv:
There was a useless 'today' variable in function's scope which is removed,
because it's already defined in top of file.


ar4s:
:ok:

szok:
:ok:

PR #436, opened 2015-04-23
Fixed pricing service costs distribution when pricing service has any services

mkurek:
When pricing service doesn't have any services, the cost was taken from all available services (empty list in if condition). Now if condition checks explicitly if `service_environments` is `None`.


ar4s:
:ok:

ghost:
:ok:

PR #438, opened 2015-04-25
fix travis coveralls

mkurek:


mkurek:
There are at least 3 ways of fixing it (and none of them is perfect):
1) move `coveralls` to script stage and run it if `make test-with-coveralls` succeeded

```
script:
  - make test-with-coveralls
  # send coveralls before processing cache (ralph packages are removed in before_cache stage and coveralls in after_success stage cannot find source files)
  - if [[ $? -eq 0 ]]; then coveralls; fi
```

EDIT: actually it looks like this method doesn't work because `make test-with-coveralls` exit code is 1 (`echo $?`), though there is info that `The command "make test-with-coveralls" exited with 0.`
2) install package once again in after_success stage

```
after_success:
  # install package again because of uninstalling it in before_cache stage
  - pip install -e .
  - coveralls
```

sometimes this may fail (to be 100% sure we have to run everything just like in install stage) - so this solution is the worst IMO
3) link checkout directory to `.../site-packages/ralph_scrooge`

```
after_success:
  - ln -s $TRAVIS_BUILD_DIR/src/ralph_scrooge `python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"`/ralph_scrooge
  - coveralls
```

Please choose the best one and I'll implement it.

PR #439, opened 2015-04-28
Fixed travis build.

ghost:
Two changes:
- added '-u root', because 'travis' user  is not visible (it's travis bug),
- fixed coveralls, by changed installation method of ralph-scrooge.


mkurek:
:ok:

ar4s:
:ok:

PR #442, opened 2015-05-22
Convert return usage values from Mongo + add simple usages do Default WH when error

damian1976:
Convert return usage values to conform to new files layout and classes division + removes the error when Warehouse does not exists when pulling Openstack simple usages (usages are added to Default WH)

### DIFF
--- a/src/ralph_scrooge/plugins/collect/_openstack_base.py
+++ b/src/ralph_scrooge/plugins/collect/_openstack_base.py
@@ -206,6 +206,7 @@ class OpenStackBasePlugin(object):
                 logger.error('Invalid warehouse: {}'.format(
                     site['WAREHOUSE']
                 ))
+                # default warehouse from fixtures
                 warehouse = Warehouse.objects.get(pk=1)
             usages = self.get_usages(today, site['CONNECTION'])
             site_new, site_total = self.save_usages(
--- a/src/ralph_scrooge/plugins/collect/_openstack_base.py
+++ b/src/ralph_scrooge/plugins/collect/_openstack_base.py
@@ -81,7 +81,7 @@ class OpenStackBasePlugin(object):
         else:
             usage_name = self.metric_tmpl.format(metric_name)
         usage_symbol = usage_name.lower().replace(':', '.').replace(' ', '_')
-        return UsageType.objects.get_or_create(
+        return UsageType.objects_admin.get_or_create(
             symbol=usage_symbol,
             defaults=dict(
                 name=usage_name,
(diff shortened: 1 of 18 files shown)


src/ralph_scrooge/plugins/cost/support.py
```diff
@@ -62,7 +62,7 @@ def _costs(
         }
         """
         logger.info("Calculating supports costs")
-        support_type = ExtraCostType.objects.get(pk=2)  # from fixture
+        support_type = ExtraCostType.objects_admin.get(pk=2)  # from fixture
```
mkurek: here only `objects` should be considered

src/ralph_scrooge/plugins/cost/team.py
```diff
@@ -216,7 +216,7 @@ def _get_cores_usage_type(self):
         """
         Physical CPU cores usage type definition
         """
-        return UsageType.objects.get_or_create(
+        return UsageType.objects_admin.get_or_create(
```
mkurek: ditto

src/ralph_scrooge/rest/allocationclient.py
```diff
@@ -48,7 +48,7 @@ def _get_service_usage_type(
     ):
         service = Service.objects.get(id=service)
         try:
-            pricing_service = PricingService.objects.get(services=service)
+            pricing_service = PricingService.objects_admin.get(services=service)
```
mkurek: only `objects`

src/ralph_scrooge/rest/allocationclient.py
```diff
@@ -241,7 +241,7 @@ def post(self, request, year, month, service, env, *args, **kwargs):
                 service__id=service,
                 environment__id=env,
             )
-            other_type = ExtraCostType.objects.get(id=1)
+            other_type = ExtraCostType.objects_admin.get(id=1)
```
mkurek: only `objects`

mkurek:
:ok:

src/ralph_scrooge/plugins/collect/_openstack_base.py
```diff
@@ -206,7 +206,7 @@ def run_plugin(self, sites, today, **kwargs):
                 logger.error('Invalid warehouse: {}'.format(
                     site['WAREHOUSE']
                 ))
-                continue
+                warehouse = Warehouse.objects.get(pk=1)
```
ghost, line 210: Add comment what is the magic '1', please.

damian1976, line 210: '1' means 'Default' warehouse

mkurek, line 210: yes - but when someone from 'outside' will look into this code he won't know this - just add some comment above this line where does this 1 come from.

ar4s:
:ok:

PR #451, opened 2015-06-13
Dedicated queue and cache for main costs recalculation process

mkurek:
- added (looking for) dedicated queue and cache (scrooge_costs_master) in main costs recalculation process
- get_queue_name and get_cache_name now accept variable number of params (*names) - first queue/cache found in settings will be returned, otherwise default


ghost:
:ok:

ar4s:
:ok:

Issue #443, opened 2015-05-22
High memory usage during costs calculation


mkurek:
"Master" job collect results from secondary level jobs, converts them to `DailyCost` instances and at the end saves everything into database (using bulk save). Creating `DailyCost` instances is unnecessary step which not only affect (time) performance but memory usage also - collected data could be transformed directly into SQL query (which also happen when using `DailyCost`) - maybe `DailyCost`could be used to automate it.

Besides that master job should be executed on separated queue (ex. `scrooge_costs_master` - actually it's executed on `scrooge_costs` with second-level jobs) to control it's destination in better way.

mkurek:
Fixed in https://github.com/allegro/ralph_pricing/pull/454

PR #454, opened 2015-06-22
Changed way of bulk creating DailyCosts to namedtuple instead of Django objects

mkurek:
- reduced memory usage while bulk creating (and saving) DailyCosts - use namedtuples insted of plain Django objects
- added additional unit tests

Connected to #443 

I've conducted some tests for memory usage and time consumption when creating `DailyCost` (or `DailyCost`-like) objects using `DailyCost` model, `AttributeDict` (dict which behave like object - you could set/get item like regular object attribute) and `namedtuple`.

I've used these 2 data structures, because django's `bulk_create` actually will work with anything supporting `getattr` (by name): https://github.com/django/django/blob/1.4.20/django/db/models/sql/subqueries.py#L156 (this is called under-the-hood when calling `bulk_create`)

`Namedtuple`s will be used in future implementation of this PR.

Next steps:
- performance tweaks of `_build_namedtuples`
- clean-up old methods (`_build_tree` and related)
- additional unit tests
- (maybe) use namedtuples in whole cost collector instead of simple dict (then there will be no need to create namedtuples in next step)

(Tests results from 1 run - just to demonstrate the order of magnitude)

Memory usage results:

| N | Django object | AttributeDict | namedtuple |
| :-: | :-: | :-: | :-: |
| 5000 | 27.7 MiB | 1.0 MiB | 0.3 MiB |
| 25000 | 145 MiB | 24.7 MiB | 18.2 MiB |
| 125000 | 739 MiB | 156 MiB | 103 MiB |

Performance results:

| N | Django object | AttributeDict | namedtuple |
| :-: | :-: | :-: | :-: |
| 5000 | 11.5s | 2.9s | 2.8s |
| 25000 | 63.9s | 14.8s | 14.6s |
| 125000 | 323s | 84s | 71s |

### DIFF
--- a/src/ralph_scrooge/models/_tree.py
+++ b/src/ralph_scrooge/models/_tree.py
@@ -5,54 +5,68 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from collections import namedtuple
+
 from django.db import models as db
+from django.db.models.base import ModelBase
 
 
-class PathFieldNotConfiguredError(Exception):
-    pass
+class NamedtupleDjangoModelMeta(ModelBase):
+    """
+    Add namedtuple mapping for Django ORM model.
+    """
+    def __new__(cls, name, bases, attrs):
+        new_class = super(NamedtupleDjangoModelMeta, cls).__new__(
+            cls, name, bases, attrs
+        )
+        # create namedtuple with fields from model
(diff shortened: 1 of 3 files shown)


andrzej-jankowski:
:ok:

ar4s:
:ok:

PR #457, opened 2015-07-31
Changed VIP collect plugin load balancer lookup and logger level

mkurek:
- Check for virtual load balancer (VirtualInfo) when asset not found
- Changed logger.warning to logger.error in VIP collect plugin when service-env does not match


ghost:
:ok:

ar4s:
:ok:

PR #458, opened 2015-09-16
[WIP] Standalone scrooge 

mkurek:
Scrooge as a standalone application! 

This is almost-working version of standalone Scrooge, without Ralph dependency (as library). I've removed ralph and ralph_assets as dependency and consecutively resolved all problem, ex. missing templates and view bases, missing user profiles, missing front-end libraries and files etc. From now Scrooge is full-featured standalone django project.

TODO (not everything in this PR of course):
Must-have:
- finish gui (not all css are moved from ralph - old-style pages are not rendering perfectly)
- make demo-data working
- adjust downloading data from ralph to NG (probably from rest api for now, maybe sth with better performance in the future)

Next steps:
- rewrite tastypie API to DRF (it'd be nice to have backward compatibility; could use django not managed models to do this easily)
- Django 1.8 & Python 3 (shouldn't be a problem)
- Dockerize

Nice to have:
- remove old-style views (all views should be in angular)
- models cleanup


ar4s:
:open_mouth:

vi4m:
Mateusz can we split work into separate iterative patches ?

vi4m:
Maybe on the separate branch if you prefer, to not break backward compatibility.

vi4m:
And port it to python 3 later

mkurek:
Yes - python 3 support (as well as django 1.9) will be definitely added later. Now i only want to get rid of ralph (2) requirement and django-bob (views). I've done it over half year now and my point of view for the standalone scrooge has changed a little bit. 

Now, old-style views (django views) are used only for reports, so maybe instead of adjusting it in the current form, we'll rewrite them to angular (new-style scrooge views) - then, scrooge will be totally independent from ralph2 UI and django-bob (it could be one PR/commit). 

In the next ones we'll break will all remaining ralph2 stuff (permissions, urls, REST API etc) - this should be quiet simple and concise.

In the last one, we'll move collect plugins from ralph "API" (API in "" on purpose, because we use it directly in python - `from ralph.util.api_scrooge import ...`) and move on to Ralph 3 REST API. 

After this 3 steps Scrooge should be completely independent from Ralph2 and able to be installed as a standalone application. Later, we could improve it by the steps I mentioned above (python 3, newest django, docker etc), but it won't be necessary to separate scrooge from ralph.

PR #459, opened 2015-09-28
OpenStack Cinder Volumes Plugin

mkurek:
Added new plugin to collect usages from OpenStack Cinder database about volumes usage.


ar4s:
:ok:

szok:
:ok:

PR #462, opened 2015-12-15
added volumes to openstack report

mkurek:
- fixed cost saving without value
- added volumes to openstack report
- save 24 instead of 1 in cinder volume when not charging for size


szok:
:ok:

ar4s:
:ok:

PR #465, opened 2016-01-20
Fix allocation admin tab

mkurek:
- don't render selects for each service-env in extra costs allocation admin tab - selects with service-env are rendered only for new rows - for existing rows text value is displayed
- fixed allocation admin reload (datepicker was not displayed before)
- fixed removing extra costs

### DIFF
--- a/src/ralph_scrooge/tests/rest/test_allocationadmin.py
+++ b/src/ralph_scrooge/tests/rest/test_allocationadmin.py
@@ -66,13 +66,13 @@ class TestAllocationAdmin(TestCase):
                             'id': 1,
                             'name': 'Other'
                         },
-                        'extra_costs': [{}]
+                        'extra_costs': []
                     }, {
                         'extra_cost_type': {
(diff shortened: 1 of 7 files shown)


ar4s:
:ok:

src/ralph_scrooge/tests/rest/test_allocationadmin.py
```diff
@@ -867,6 +867,56 @@ def test_update_extra_cost_when_everything_is_ok(self):
...
+                        'cost': cost,
+                        'forecast_cost': forecast_cost,
+                        'service': service_environment.service_id,
+                        'env': service_environment.environment_id,
+                    }]
+                }]
+            },
+            format='json'
+        )
+        self.assertEqual(models.ExtraCost.objects.count(), 1)
```
ghost, line 918: `self.assertEqual(models.ExtraCost.objects[0], extra_cost.id)`?

ghost:
:ok:

PR #473, opened 2016-08-11
[switch to R3] Base for other plugins (querying Ralph, model changes etc.).

xor-xor:
From all PRs marked as "switch to R3", this one should be merged first (all the rest are based on it).


src/ralph_scrooge/models/warehouse.py
```diff
@@ -31,6 +31,12 @@ class Warehouse(TimeTrackable, EditorTrackable, Named,
         blank=False,
         unique=True,
     )
+    r3_id_from_assets = db.IntegerField(
```
mkurek: name it just `r3_id`

mkurek:
:ok:

PR #476, opened 2016-08-11
[switch to R3] Add ralph3_asset_model plugin.

xor-xor:

BTW, there were no tests for the original plugin (i.e. Ralph2-based one).


mkurek:
:ok:

PR #477, opened 2016-08-11
[switch to R3] Add ralph3_data_center plugin.

xor-xor:

There were no tests for the original (i.e. Ralph2-based) plugin.

### DIFF
--- a/src/ralph_scrooge/plugins/collect/ralph3_data_center.py
+++ b/src/ralph_scrooge/plugins/collect/ralph3_data_center.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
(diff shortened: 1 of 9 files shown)


src/ralph_scrooge/plugins/collect/ralph3_data_center.py
```diff
@@ -0,0 +1,44 @@
...
+from ralph_scrooge.models import Warehouse
+from ralph_scrooge.plugins.collect.utils import get_from_ralph
+
+
+logger = logging.getLogger(__name__)
+
+
+def update_data_centers(dc_from_ralph):
+    warehouse, created = Warehouse.objects.get_or_create(
+        id_from_assets=dc_from_ralph['id']
```
mkurek: what about new column for Ralph3 id?

mkurek:
:ok:

PR #482, opened 2016-08-19
WIP Add monthly costs to rest API

szok:
TODO: 
- Tests

### DIFF
--- a/src/ralph_scrooge/rest/monthly_costs.py
+++ b/src/ralph_scrooge/rest/monthly_costs.py
@@ -1,0 +1,257 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import logging
+import time
+from dateutil import rrule
+
+from django.conf import settings
+from django.core.cache import get_cache
+from django.db import transaction
+from django.utils.translation import ugettext_lazy as _
+from rq import get_current_job
+
+from rest_framework import status
+from rest_framework.views import APIView
+from rest_framework.response import Response
+
+from ralph_scrooge.plugins.cost.collector import Collector
+from ralph_scrooge.utils.common import get_cache_name, get_queue_name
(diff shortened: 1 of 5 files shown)


src/ralph_scrooge/rest/monthly_costs.py
```diff
@@ -0,0 +1,257 @@
...
+from ralph_scrooge.forms import MonthlyCostsForm
+from ralph_scrooge.models import CostDateStatus
+
+
+logger = logging.getLogger(__name__)
+
+
+class MonthlyCosts(APIView, WorkerJob):
+
+    Form = MonthlyCostsForm
```
mkurek, line 34: do we have to use django (gui) form here? couldn't we use DRF serializer?

src/ralph_scrooge/rest/monthly_costs.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def get(self, *args, **kwargs):
+        if self.request.GET:
+            self.form = self.Form(self.request.GET)
+        else:
+            self.form = self.Form()
+
+        if self.form.is_valid():
+            result = {}
+            if 'recalculate' in self.request.GET:
```
mkurek, line 50: could we make it more restful? sth like:
- i send POST request with start date, end date and info about forecast (fields like in MonthlyCostsForm)
- i receive 202 response with details url in some header (Location?) - job id or md5 of params or sth like that could be used to build url
- i do GET request on returned url to check for updates

src/ralph_scrooge/rest/monthly_costs.py
```diff
@@ -0,0 +1,257 @@
...
+                if self.progress < 100:
+                    result['result'] = _(
+                        'Please wait for costs recalculation.'
+                    )
+                else:
+                    result['result'] = _('Costs recalculated.')
+                    result['data'] = sorted(
+                        self.data.items(), key=lambda k: k[0]
+                    )
+            elif 'accept_costs' in self.request.GET:
```
mkurek, line 65: it should be separated endpoint IMO

PR #485, opened 2016-08-26
Ralph3 sync plugins improvements

mkurek:
Migrator for AssetModel, BusinessSegment, DataCenter, Environment and ProfitCenter from CSV files from Ralph3.

Fixed current Ralph3 plugins.


szok:
:ok:

PR #486, opened 2016-08-29
Update for cloud project Ralph3 collect plugin

mkurek:
- store tenant_id
- rename plugin (cloud_project -> ralph3_cloud_project)
- log tenant id when invalid service env


szok:
:ok:

xor-xor:
:ok:

PR #487, opened 2016-08-30
Add Ralph3 plugins mapping (they replace some of Ralph2 plugins)

mkurek:
Mark Ralph2-related plugins as done when processing their Ralph3-equivalent to enable processing of not-rewrited plugins (which depends on Ralph2-related plugins, like asset, openstack\* etc).


szok:
:ok:

xor-xor:
:ok:

PR #492, opened 2016-09-02
Handle MultipleObjectsReturned in tenant, asset and vip plugins

mkurek:
After adding new column "ralph3_id" etc, there might be multiple rows with None id from Ralph2.

### DIFF
--- a/src/ralph_scrooge/plugins/collect/asset.py
+++ b/src/ralph_scrooge/plugins/collect/asset.py
@@ -168,12 +168,15 @@ def update_assets(data, date, usages):
             service__ci_id=data['service_id'],
             environment__ci_id=data['environment_id'],
         )
-    except ServiceEnvironment.DoesNotExist:
+    except (
+        ServiceEnvironment.DoesNotExist,
+        ServiceEnvironment.MultipleObjectsReturned
+    ):
         raise ServiceEnvironmentDoesNotExistError()
 
     try:
         warehouse = Warehouse.objects.get(id_from_assets=data['warehouse_id'])
-    except Warehouse.DoesNotExist:
+    except (Warehouse.DoesNotExist, Warehouse.MultipleObjectsReturned):
         warehouse = Warehouse.objects.get(pk=1)  # Default from fixtures
 
     asset_info, new_created = get_asset_info(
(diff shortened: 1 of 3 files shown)


szok:
:ok:

src/ralph_scrooge/plugins/collect/asset.py
```diff
@@ -168,12 +168,12 @@ def update_assets(data, date, usages):
             service__ci_id=data['service_id'],
             environment__ci_id=data['environment_id'],
         )
-    except ServiceEnvironment.DoesNotExist:
+    except (ServiceEnvironment.DoesNotExist, ServiceEnvironment.MultipleObjectsReturned):
         raise ServiceEnvironmentDoesNotExistError()
```
xor-xor, line 175: I'm wondering if the name of this exception is still valid in the above circumstances (does not exist vs. multiple objects returned).

EDIT: maybe we should handle these two cases separately..?

mkurek: i don't want to spend too much time on this plugin, since it'll be removed probably next week - it's just temporary workaround to make it working after adding ralph3 columns.

src/ralph_scrooge/plugins/collect/tenant.py
```diff
@@ -42,7 +42,7 @@ def save_tenant_info(ralph_tenant, unknown_service_environment):
             service__ci_id=ralph_tenant['service_id'],
             environment__ci_id=ralph_tenant['environment_id'],
         )
-    except ServiceEnvironment.DoesNotExist:
+    except (ServiceEnvironment.DoesNotExist, ServiceEnvironment.MultipleObjectsReturned):
```
xor-xor: I'd consider handling these two cases separately, or changing the logger message below.

src/ralph_scrooge/plugins/collect/vip.py
```diff
@@ -48,7 +48,7 @@ def save_vip_info(ralph_vip, unknown_service_environment):
             environment__ci_id=ralph_vip['environment_id'],
         )
         service_environment_found = True
-    except ServiceEnvironment.DoesNotExist:
+    except (ServiceEnvironment.DoesNotExist, ServiceEnvironment.MultipleObjectsReturned):
```
xor-xor: See previous comment.

src/ralph_scrooge/plugins/collect/asset.py
xor-xor: OK, cool. So ignore the rest of my comments in this PR :wink:

xor-xor:
:ok:

PR #493, opened 2016-09-02
Move plugin handler from Ralph2 to Scrooge

mkurek:
Logic of plugins (chaining, running next available etc) copied from ralph.util.plugin.


src/ralph_scrooge/plugins/plugin.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python
```
xor-xor: What about renaming this file from `plugin.py` to `plugin_runner.py`..? `from ralph_scrooge.plugins import plugin` looks kind of strange to me, especially if such import is placed in the actual plugin :wink:

src/ralph_scrooge/plugins/plugin.py
```diff
@@ -0,0 +1,107 @@
...
+BY_REQUIREMENTS = {}
+PRIORITIES = {}
+
+
+class PluginFailed(Exception):
+
+    """Raised when a scan plugin is unsuccessful."""
+
+
+class Restart(Exception):
```
xor-xor: What about renaming it to something more descriptive, e.g. `RestartNeeded` or `RestartNeededException` (or something similar)..?

src/ralph_scrooge/plugins/plugin.py
```diff
@@ -0,0 +1,107 @@
...
+something more object-oriented.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+BY_NAME = {}
+BY_REQUIREMENTS = {}
```
xor-xor: What about `PLUGINS_BY_NAME` and `PLUGINS_BY_REQUIREMENTS`..?

src/ralph_scrooge/plugins/plugin.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    Run a single plugin by a name.
+    """
+    return BY_NAME[chain][func_name](**kwargs)
+
+
+def purge(plugin_set):
+    """Remove all plugins from the plugin subsystem which are not related by
+    means of requirements to the given `plugin_set`."""
+    for ck, cv in BY_REQUIREMENTS.iteritems():
```
xor-xor: I have nothing against such short names for variables used in loops (IMHO it's actually good idea),  but what does `c` mean here (in `ck`, `cv`)..? I guess that by changing `BY_REQUIREMENTS` to something more descriptive, it would be clear enough to see that.

src/ralph_scrooge/plugins/plugin.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def run(chain, func_name, **kwargs):
+    """
+    Run a single plugin by a name.
+    """
+    return BY_NAME[chain][func_name](**kwargs)
+
+
+def purge(plugin_set):
```
xor-xor: I'm just wondering, if changing all these function names  to `*_plugin` or `*_plugins` (e.g., `run_plugin`, `purge_plugins` etc.) would be a good idea - what do you think?

src/ralph_scrooge/plugins/plugin.py
```diff
@@ -0,0 +1,107 @@
...
+        {},
+    ).setdefault(
+        frozenset(requires),
+        [],
+    ).append(func.func_name)
+    PRIORITIES.setdefault(chain, {})[func.func_name] = priority or 100
+    return func
+
+
+def next(chain, done_reqs):
```
xor-xor: IMO `next` implies singular, while the docstring says plural (i.e. "plugins").

src/ralph_scrooge/plugins/plugin.py
```diff
@@ -0,0 +1,107 @@
...
+        requires = set()
+    BY_NAME.setdefault(chain, {})[func.func_name] = func
+    BY_REQUIREMENTS.setdefault(
+        chain,
+        {},
+    ).setdefault(
+        frozenset(requires),
+        [],
+    ).append(func.func_name)
+    PRIORITIES.setdefault(chain, {})[func.func_name] = priority or 100
```
xor-xor: I'd add some comment explaining what is the highest possible priority here.

szok:
:ok:

xor-xor:
:ok:

PR #496, opened 2016-09-07
Add HTTP403 view from ralph

szok:


src/ralph_scrooge/utils/views.py
```diff
@@ -0,0 +1,31 @@
...
+import functools
+
+from django.http import (
+    HttpResponse,
+    HttpResponseRedirect,
+    HttpResponseNotAllowed
+)
+
+
+def jsonify(func):
```
mkurek: Handled in #497 - remove it

mkurek:
change it to DRF 403

PR #502, opened 2016-09-08
renamed vipinfo migration from 0025 to 0026

mkurek:
there was a conflict between 0025 migrations (virtual and vip)


xor-xor:
:ok:

szok:
:ok:

PR #504, opened 2016-09-08
Set lowest possible port for VIP to 0

mkurek:
For some specific load balancers types, min port could be 0 (and it could be treated as "wildcard")


xor-xor:
:ok:

szok:
:ok:

PR #507, opened 2016-09-12
Migrate API to DjRF

xor-xor:
The purpose of this PR is to switch `pricingserviceusages` API endpoint do Django Rest Framework (from TastyPie). The structure of data that can be sent/fetched to/from this endpoint didn't change in order to keep backward compatibility.

This PR also adds custom authentication scheme `TastyPieLikeTokenAuthentication`, which extends DjRF's `TokenAuthentication` with support of `Authorization` header in the format that is expected by TastyPie.

It also gives `/scrooge/api-token-auth/` endpoint for generation of DjRF's API keys.

To facilitate TastyPie -> DjRF migration, `scrooge_migrate_api_keys` management command is provided, which copies API keys that already have been generated by TastyPie into a structure used by DjRF.

**Important note:** on Ralph's side, this PR requires update of `South` package to version `1.0.2`, adding `rest_framework.authtoken` to `INSTALLED_APPS`, and performing `ralph syncdb` followed by `ralph migrate rest_framework.authtoken`.


src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,335 @@
...
+
+class PricingServiceUsageObjectSerializer(Serializer):
+    pricing_service = serializers.CharField()
+    pricing_service_id = serializers.IntegerField()
+    date = serializers.DateField()
+    overwrite = serializers.CharField()
+    usages = UsagesObjectSerializer(many=True)
+
+    class Meta:
+        model = PricingServiceUsageObject
```
mkurek: do we need model here? imo it's not `ModelSerializer`, so there is no need for specifying model here (and we don't need this `PricingServiceUsageObject` etc model either - we could operate on python structures, like dict)

src/ralph_scrooge/rest/urls.py
```diff
@@ -108,4 +109,20 @@
...
+    # XXX To be eventually moved into `/api/` URLs hierarchy.
+    url(
+        r'^pricingserviceusages/?$',
+        # scrooge_permission(PricingServiceUsages.as_view()),
+        PricingServiceUsages.as_view(),
+        name='pricing_service_usages'
+    ),
+    url(
+        # r'^pricingserviceusages/?$',
+        r'^pricingserviceusages/(?P<pricing_service_id>[\w\d\s_.-]+)/(?P<date>[\d-]+)/$',
```
mkurek: pricing service id could only be number

mkurek: could you make some extended regex for date (to match YYYY-MM-DD)?

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,335 @@
...
+    #     # assert isinstance(post_data['date'], datetime.date)  # XXX
+
+    #     # check if overwrite is properly set
+    #     assert post_data.get('overwrite', 'no') in (  # XXX
+    #         'no',
+    #         'delete_all_previous',
+    #         'values_only',
+    #     )
+
+    def get(self, request, *args, **kwargs):
```
mkurek: couldn't `date` and `pricing_service_id` be specified here, as method params?

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,335 @@
...
+            kwargs['date'],
+            '%Y-%m-%d'
+        ).date()
+        pricing_service_id = kwargs['pricing_service_id']
+        result = self.get_as_obj(usages_date, pricing_service_id)
+        return Response(PricingServiceUsageObjectSerializer(result).data)
+
+    def post(self, request, *args, **kwargs):
+        post_data = json.loads(request.raw_post_data)  # XXX what about using BytesIO/JSONParser for that..?
+        deserializer = PricingServiceUsageObjectDeserializer(data=post_data)
```
mkurek: isn't `deserializer = PricingServiceUsageObjectDeserializer(data=request.data)` working?

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,335 @@
...
+        if deserializer.is_valid():
+            self.save_usages(post_data)
```
mkurek: why not passing serializer here?

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,335 @@
...
+            ]):
+                pass
+                # XXX raise some exception here
+            else:
+                service_env = self.get_service_env(usages)
+                pricing_object = service_env.dummy_pricing_object  # XXX what is dummy_pricing_object?
+        return pricing_object
+
+    def get_usages(self, post_data):
+        usage_types = {}  # XXX there's such empty dict in old API's code
```
mkurek: this could be moved to (cached) method to get usage type by symbol

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,335 @@
...
+                previuos_usages = DailyUsage.objects.filter(date=date, type=k)
+                if overwrite == 'values_only':
+                    previuos_usages = previuos_usages.filter(
+                        daily_pricing_object__in=v
+                    )
+                previuos_usages.delete()
+
+    def get_pricing_object(self, usages):
+        if usages.get('pricing_object'):
+            pricing_object = PricingObject.objects.get(
```
mkurek: i don't know if there is need for try-except - it should be handled at validation level - imo validation could already fetch right objects (UsageType, PricingObject, ServiceEnvironment etc) to this dict

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done.

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done.

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: You mean:
`self.save_usages(deserializer)`
or:
`self.save_usages(deserializer.object)`
?

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: As discussed face-to-face - we won't be touching this now.

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: I'll try dict-based approach here as well, and see which one will be better.

src/ralph_scrooge/rest/urls.py
xor-xor: Done.

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,369 @@
...
+        'service_id': service_id,
+        'environment': environment,
+        'pricing_object': pricing_object,
+        'usages': usages or [],
+    }
+
+
+class UsagesSerializer(Serializer):
+    service = serializers.CharField()
+    service_id = serializers.IntegerField()
```
mkurek, line 114: you could add `service_uid` field here too

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,369 @@
...
+    return {
+        'service': service,
+        'service_id': service_id,
+        'environment': environment,
+        'pricing_object': pricing_object,
+        'usages': usages or [],
+    }
+
+
+class UsagesSerializer(Serializer):
```
mkurek, line 112: couldn't we (re)use `UsagesDeserializer` for both serialization and deserialization? since we're serializing objects, we don't need `required` params or it could be set to False - it's only relevant to the user data

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,369 @@
...
+        }
+        if attrs.get('service_id'):
+            se_params['service_id'] = attrs['service_id']
+        elif attrs.get('service_uid'):
+            se_params['service__ci_uid'] = attrs['service_uid']
+        else:
+            se_params['service__name'] = attrs['service']
+        try:
+            se = ServiceEnvironment.objects.get(**se_params)
+        except ServiceEnvironment.DoesNotExist:
```
mkurek, line 152: there could be the case when multiple `ServiceEnvironment`s will be returned as well (ex. when user provides name of the service)

doc/api.rst
```diff
@@ -1,85 +1,115 @@
...
     "date": "<date>",
-    "overwrite: "no|values_only|delete_all_previous",
+    ["overwrite: "no | values_only | delete_all_previous"],
     "usages": [
         {
-            "service": "<service name or uid or id>",
-            "pricing_object": "<pricing_object_name>",
+            "pricing_object": "<pricing_object_name>" |
+            (
+              ("service": "<service name>" | "service_id": <service ID> | "service_uid": "<service UID>"),
```
mkurek: service ID -> service ID (in Scrooge)

doc/api.rst
```diff
@@ -1,85 +1,115 @@
...
-  Send a JSON message to API endpoint (``RALPH_URL/scrooge/api/v0.9/pricingserviceusages/``) using POST method with data in the format described in the “Technical description” section.
+-------------------
+Pushing usages data
+-------------------
 
+Send your usages data to ``/scrooge/api/v0.9/pricingserviceusages/``
+endpoint, using POST method. Your data should be in JSON format, with
+the structure described below. Authorization is done with
+``Authorization`` HTTP header containing your user name and API key
+generated in Scrooge's Admin.
```
mkurek: provide some example in this section too (i know it's later in example, but imo `Authorization: ApiKey <user name>:<API key from Scrooge Admin>` should be placed here, and later in example we should operate on "real" data, like `Authorization: ApiKey my_technical_user: 8356dd1592c789e946039774e00f72019c93c5ea`

doc/api.rst
```diff
@@ -156,49 +197,124 @@ Example::
...
 """""""""""""""""""""""""""
 Overwriting previous values
 """""""""""""""""""""""""""
 
-API provides a way to define how to treat previous service usages values uploaded for a given date (e.g. when data is sent twice for a given date). The possible actions (overwrite) are:
+The aforementioned ``overwrite`` field defines a way how to treat
+previous service usage values uploaded for the same date and usage
+type. There are three possible actions here:
+
+* ``delete_all_previous`` - all previously upladed daily usages for
```
mkurek: uploaded

doc/api.rst
```diff
@@ -156,49 +197,124 @@ Example::
...
+    daily usage 2: service env 2, value 60
+
+    2nd upload (same day, same usage type):
+    daily usage 1: service env 1, value 50
+
+    final result:
+    daily usage 1: service env 1, value 50
+
+* ``values_only`` - all previously uploaded daily usages from the same
+  date, with the same usage type *and the same service environment*
```
mkurek, line 227: service env or pricing object

doc/api.rst
```diff
@@ -156,49 +197,124 @@ Example::
...
 
 """""""""""""""""""""""""""""""
 Possible responses (HTTP codes)
 """""""""""""""""""""""""""""""
 
-201 - everything ok, data saved properly.
+201 - everything OK, data is saved successfully.
 
-400 - invalid symbol or name (ex. usage type or service) - there will be provided detailed information about invalid data.
+400 - invalid symbol or name (e.g. usage type or service) - such error
```
mkurek: maybe invalid data in general?

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,430 @@
...
+# (especially for usages).
+
+# TODO(xor-xor) Consider separate endpoints for usages associated with:
+# * service + environment
+# * service_id + environment
+# * service_uid + environment
+# * pricing_object
+# instead of a "polymorphic" one
+
+# TODO(xor-xor): Consider json-schema based approach instead of all of those
```
mkurek, line 54: 👍

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,430 @@
...
+        try:
+            PricingService.objects.get(name=value)
+        except PricingService.DoesNotExist:
+            raise serializers.ValidationError(
+                "Unknown service name: {}".format(value)
+            )
+        return attrs
+
+
+class TastyPieLikeTokenAuthentication(TokenAuthentication):
```
mkurek, line 282: could you move it to other (dedicated) module?

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,430 @@
...
+@authentication_classes((TastyPieLikeTokenAuthentication,))
+@permission_classes((IsAuthenticated,))
+def list_pricing_service_usages(
+        request, usages_date, pricing_service_id, *args, **kwargs
+):
+    try:
+        PricingService.objects.get(id=pricing_service_id)
+    except PricingService.DoesNotExist:
+        msg = (
+            "Service with ID {} does not exist.".format(pricing_service_id)
```
mkurek: Service -> Pricing Service

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,430 @@
...
+        )
+        return Response({'error': msg}, status=status.HTTP_400_BAD_REQUEST)
+    usages = get_usages(usages_date, pricing_service_id)
```
mkurek: you could store result of `PricingService.objects.get(id=pricing_service_id)` in some variable and pass it to `get_usages`

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,430 @@
...
+    for dpo, u in usages_dict.iteritems():
+        se = dpo.service_environment
+        usages = new_usages(
+            service=se.service.name,
+            service_id=se.service.id,
+            environment=se.environment.name,
+        )
+        # If pricing object is not a dummy pricing object for given service
+        # environment, then use it.
+        if se.dummy_pricing_object != dpo.pricing_object:
```
mkurek: we could check dpo.pricing_object.type here (and compare it with PRICING_OBJECT_TYPES.DUMMY) - it should be faster (no need to fetch dummy pricing object for service env)

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,430 @@
...
+    )
+    remove_previous_daily_usages(
+        pricing_service_usage['overwrite'],
+        pricing_service_usage['date'],
+        usages_daily_pricing_objects,
+    )
+    DailyUsage.objects.bulk_create(daily_usages)
+
+
+def get_usages_for_save(pricing_service_usage):
```
mkurek, line 432: could you add some docstrings for not-obvious methods here (longer than few lines)?

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,430 @@
...
+    usage_types_cache = {}
+    daily_usages = []
+    usages_daily_pricing_objects = defaultdict(list)
+
+    for usages in pricing_service_usage['usages']:
+        for usage in usages['usages']:
+            usage_type = usage_types_cache.get(
+                usage['symbol'],
+                UsageType.objects.get(symbol=usage['symbol'])
```
mkurek: it'll not work as a cache - python will evaluate `UsageType.objects.get(symbol=usage['symbol'])` before fetching item from dict - use maybe try-except instead

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -0,0 +1,430 @@
...
+@permission_classes((IsAuthenticated,))
+def create_pricing_service_usages(request, *args, **kwargs):
+    deserializer = PricingServiceUsageDeserializer(data=request.DATA)
+    if deserializer.is_valid():
+        save_usages(deserializer.object)
+        return HttpResponse(status=201)
+    return Response(deserializer.errors, status=400)
+
+
+def save_usages(pricing_service_usage):
```
mkurek, line 411: wrap it in db transaction (decorator)

mkurek (changes requested):

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: I'd leave them as separate entities, for clarity - I think that the lack of symmetry here, while having it other cases (i.e. `UsageSerializer` / `UsageDeserializer`, `PricingServiceUsageSerializer` / `PricingServiceUsageDeserializer`) may be confusing. In other words, I prefer consistency even if it costs extra few lines of code :wink:

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: OK, so I'll leave this TODO here and come back for it when we finally separate Scrooge from Ralph (I'll create a ticket for it in our JIRA as well).

src/ralph_scrooge/rest/pricing_service_usages.py
mkurek: ok

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done (although keep in mind, that this field wasn't a part of the original API, so it may cause problems on receivers' part).

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done (such case will result in validation error).

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: I'm not sure if such move makes sense now - i.e. it's just one, small class which is only used here. I'd consider a separate module for this when we would have some other, cutom auth scheme or if this scheme would be used in some other places as well.

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done.

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done.

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: You mean `django.db.transaction.commit_on_success` or something else..?

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done.

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done.

doc/api.rst

doc/api.rst
xor-xor: Done.

doc/api.rst

doc/api.rst

doc/api.rst

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done.

src/ralph_scrooge/rest/pricing_service_usages.py
mkurek: yes

src/ralph_scrooge/rest/pricing_service_usages.py
mkurek: ok

src/ralph_scrooge/rest/pricing_service_usages.py
xor-xor: Done.

mkurek (approved):

PR #508, opened 2016-09-13
Fix parsing end date in usages report

mkurek:
Take end date instead of start (before, report was generated for one-day-only)


szok:
:ok:

PR #514, opened 2016-09-28
Fix PricingObjectModel model_id field migration (nullability)

mkurek:
Somehow (?) between migrations 0007 and 0008, model_id of PricingObjectModel changed from beign not-null to null and it was reflected in south models dump, but there was no migration action making it nullable in database. This commit fixes it by removing info about null and blank from south models dump in the latest migration and creating new migration which make this field nullable in db.


szok (approved):

xor-xor:
:ok:

PR #516, opened 2016-09-29
Add endpoint for uploading team time divisions.

xor-xor:


doc/api.rst
```diff
@@ -334,6 +336,119 @@ Example::
...
+      },
+      {
+        "service_uid": "uid-444",
+        "environment": "prod",
+        "percent": 10.0
+      }
+    ]
+  }
+
+Here, we can see that our team (given in URL by its ID, i.e. ``156``)
```
mkurek: change it to `123` to be the same as in the example

doc/api.rst
```diff
@@ -334,6 +336,119 @@ Example::
...
+        "environment": "prod",
+        "percent": 10.0
+      }
+    ]
+  }
+
+
+.. note:: ``teamtimedivision`` endpoint is only accessible to team
+   managers (both for read and write). A team manager in Scrooge's is
+   a person who is associated with a team, and is an owner of at least
```
mkurek: are you sure about "and is an owner of at least one service"? I'm not :)

src/ralph_scrooge/rest/team_time_division.py
```diff
@@ -0,0 +1,228 @@
...
+)
+from ralph_scrooge.rest.common import get_dates
+
+
+class PercentSerializer(Serializer):
+    service_uid = serializers.CharField(required=True)
+    environment = serializers.CharField(required=True)
+    percent = serializers.FloatField(required=True)
+
+    def validate_service_uid(self, attrs, source):
```
mkurek: imo we don't have to validate service_uid and env name separately - ServiceEnvioronment validation (like in `validate` method) will be enough (and we'll have 3x less db queries).

src/ralph_scrooge/rest/team_time_division.py
```diff
@@ -0,0 +1,228 @@
...
+        'service_uid': service_uid,
+        'environment': environment,
+        'percent': percent,
+    }
+
+
+def _get_percents(team, start, end):
+    percents = []
+    for tsep in TeamServiceEnvironmentPercent.objects.filter(
+        team_cost__team__id=team,
```
mkurek: `team_cost__team=team` (join for matching by id is not necessary)

src/ralph_scrooge/rest/team_time_division.py
```diff
@@ -0,0 +1,228 @@
...
+    def validate(self, attrs):
+        err = None
+        division = attrs.get('division')
+        if division is None:
+            return attrs
+
+        # validate percents (if they sum up to 100)
+        percent_total = 0
+        for d in division:
+            percent_total += d['percent']
```
mkurek, line 101: what if someone does not provide percent key or it's not a number? at which point is it validated? could we get there if there is some data corrupted earlier?

src/ralph_scrooge/rest/team_time_division.py
```diff
@@ -0,0 +1,228 @@
...
+        if percent_total != 100:
+            err = (
+                "Percents should sum to 100, now it's {}."
+                .format(percent_total)
+            )
+
+        # validate uniqueness of service_uid/environment per single division
+        uids_and_envs = [
+            "{}/{}".format(
```
mkurek, line 110: couldn't you use tuple here - `(d['service_uid'], d['environment'])`?

src/ralph_scrooge/rest/team_time_division.py
```diff
@@ -0,0 +1,228 @@
...
+    return {
+        'team_id': team_id,
+        'year': year,
+        'month': month,
+        'division': division or [],
+    }
+
+
+class TeamTimeDivision(APIView):
+    authentication_classes = (TastyPieLikeTokenAuthentication,)
```
mkurek, line 140: couldn't we force regular DRF authentication here?

src/ralph_scrooge/rest/team_time_division.py
```diff
@@ -0,0 +1,228 @@
...
+    permission_classes = (IsAuthenticated, IsTeamLeader)
+
+    def get(self, request, year, month, team_id, *args, **kwargs):
+        year, month, team_id = _args_to_int(year, month, team_id)
+        try:
+            Team.objects.get(id=team_id)
+        except Team.DoesNotExist:
+            err = "Team with ID {} does not exist.".format(team_id)
+            return Response(
+                {'error': err}, status=status.HTTP_400_BAD_REQUEST
```
mkurek: shouldn't we return 404 here?

src/ralph_scrooge/rest/team_time_division.py
```diff
@@ -0,0 +1,228 @@
...
+        team_id=team_id,
+        start=first_day,
+        end=last_day,
+    )[0]
+    for percent in division:
+        service_environment = ServiceEnvironment.objects.get(
+            service__ci_uid=percent.get('service_uid'),
+            environment__name=percent.get('environment')
+        )
+        tsep, created = TeamServiceEnvironmentPercent.objects.get_or_create(
```
mkurek: do we need to use `get_or_create` here? since we're deleting all previous values before, and validating uniqueness of the data (service-env), imo we could just simply create new records.

mkurek (changes requested):

doc/api.rst

doc/api.rst
xor-xor: Right. I've changed this sentence to: "A team manager in Scrooge's is a person who is associated with a given team, and is marked as an owner of some (zero or more) services [...]", although "being an owner of zero services" sounds kind of strange to me ;) Maybe we should completely drop that part re: ownership..?

src/ralph_scrooge/rest/team_time_division.py
xor-xor: ...and lose the ability to authenticate users with old API keys (i.e., the ones from TastyPie)..?

EDIT: I mean, I know that at this point all old API keys _should_ be migrated to DjRF's format (with `scrooge_migrate_api_keys` mgmt command), but can we safely assume, that all Scrooge installations went through this step already..?

src/ralph_scrooge/rest/team_time_division.py
xor-xor: Right. I was influenced by [this piece of code](https://github.com/xor-xor/ralph_pricing/blob/b275031b762e440db1d715258b880d0ab597e334/src/ralph_scrooge/rest/allocationclient.py#L325).

xor-xor: Done.

src/ralph_scrooge/rest/team_time_division.py
xor-xor: Done

src/ralph_scrooge/rest/team_time_division.py
xor-xor: I could, but I'd stay with strings, because we re-use them when the error message is being constructed.

src/ralph_scrooge/rest/team_time_division.py
xor-xor: Done.

src/ralph_scrooge/rest/team_time_division.py
xor-xor: Done.

src/ralph_scrooge/rest/team_time_division.py
xor-xor: ad. 1. You'll get "This field is required."
ad. 2. You'll get "value must be a float."
ad. 3. [Here](url)..? :wink: 
ad. 4. Could you expand "some data corrupted earlier"..?

doc/api.rst
mkurek: Maybe "A team manager in Scrooge's is a person who is associated with a given team and is allowed to allocate team resources (time) to services"?

mkurek (approved):

szok (approved):

PR #520, opened 2016-09-30
Change cost collector to static class

ar4s:


mkurek (approved):

xor-xor:
I'm just wondering...  Coulnd't `Collector` be just a simple module..? It's a singleton, there's no state shared between its methods and nothing inherits from it - so what's the purpose of it to be a class in the first place..?

xor-xor:
Closing this by now - we can always revisit ideas/suggestions presented here anyway.

PR #525, opened 2016-10-04
Upgrade django version from 1.4 to 1.5

szok:
Upgrade django version from 1.4 to 1.5 and migration user profile to user model


xor-xor:
Code-wise seems to be OK, but to finally approve this PR, I need to check if Scrooge works properly after those changes - I'll let you know :wink: 

EDIT: Functionality-wise seems to be OK as well - I've been clicking through GUI and I didn't notice any errors or strange behaviors. The only thing that I've got is a DB error related to deleteion of `auth_user` table during `scrooge syncdb`, but I guess that it can be ignored..?

xor-xor (approved):

src/ralph_scrooge/admin.py
```diff
@@ -31,6 +33,31 @@ def get_readonly_fields(self, request, obj=None):
         return self.readonly_fields
 
 
+class ScroogeUserChangeForm(UserChangeForm):
+    class Meta:
+        model = models.ScroogeUser
+
+
+@register(models.ScroogeUser)
+class ScroogeUserAdmin(UserAdmin):
```
mkurek, line 42: couldn't we use default django admin for user here?

src/ralph_scrooge/models/owner.py
```diff
@@ -5,14 +5,54 @@
...
-from django.contrib.auth.models import User
+from django.contrib.auth.models import AbstractUser
 from django.db import models as db
 from django.utils.translation import ugettext_lazy as _
 
 from ralph_scrooge.utils.models import TimeTrackable
 from dj.choices import Choices, Country, Gender
 
 
+class ScroogeUser(AbstractUser):
```
mkurek, line 16: imo we don't need nick, birth_date, gender, city, activation_token, department, manager, location, segment, company fields in scrooge. Actually, we could remove all of these fields :)

src/ralph_scrooge/models/owner.py
```diff
@@ -5,14 +5,54 @@
...
+    department = db.CharField(max_length=64, blank=True)
+    manager = db.CharField(max_length=1024, blank=True)
+    location = db.CharField(max_length=128, blank=True)
+    segment = db.CharField(max_length=256, blank=True)
+
+    class Meta:
+        verbose_name = _('user')
+        verbose_name_plural = _('users')
+        app_label = 'ralph_scrooge'
+        db_table = 'scrooge_user'
```
mkurek: do we need this?

mkurek (changes requested):

src/ralph_scrooge/admin.py
szok: I use this to add additional fields. If we don't need it I will remove this class

src/ralph_scrooge/models/owner.py
szok: ok

src/ralph_scrooge/models/owner.py
szok: Yes, without this did not work

src/ralph_scrooge/migrations/0030_copy_userprofile_to_scrooge_user.py
```diff
@@ -0,0 +1,520 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
```
mkurek: shouldn't it be DataMigration? http://south.readthedocs.io/en/latest/tutorial/part3.html#data-migrations

src/ralph_scrooge/migrations/0030_copy_userprofile_to_scrooge_user.py
```diff
@@ -0,0 +1,520 @@
...
+            user_id = old_user_dict.pop('id')
+            username = old_user_dict.pop('username')
+            del old_user_dict['_state']
+            new_user, _ = orm['ralph_scrooge.ScroogeUser'].objects.get_or_create(
+                id=user_id,
+                username=username,
+                defaults=old_user_dict
+            )
+            new_user.save()
+            Owner = orm['ralph_scrooge.Owner']
```
mkurek, line 23: couldn't we get rid of `Owner` model here? we could use `ScroogeUser` directly instead.

src/ralph_scrooge/models/owner.py
```diff
@@ -65,30 +75,34 @@ class OwnershipType(Choices):
 class OwnerManager(db.Manager):
     def get_query_set(self):
         return super(OwnerManager, self).get_query_set().select_related(
-            'profile',
-            'profile__user'
+            'user'
         )
 
 
 class Owner(TimeTrackable):
```
mkurek, line 82: as I said earlier, everything pointing to Owner (ForeignKey to Owner) could be changed to ForeignKey to ScroogeUser

src/ralph_scrooge/admin.py
```diff
@@ -31,6 +33,31 @@ def get_readonly_fields(self, request, obj=None):
...
+    fieldsets = (
+        (None, {'fields': ('username', 'password')}),
+        (_('Personal info'), {'fields': ('first_name', 'last_name', 'email')}),
+        (_('Permissions'), {'fields': ('is_active', 'is_staff', 'is_superuser',
+                                       'groups', 'user_permissions')}),
+        (_('Important dates'), {'fields': ('last_login', 'date_joined')}),
+        (_('Profile'), {
+            'fields': [
+                'nick', 'birth_date', 'gender', 'country', 'city', 'company',
+                'employee_id', 'profit_center', 'cost_center', 'department',
```
mkurek: you've removed these fields from model - remove them here too (or even use plain django model admin for user)

mkurek (changes requested):

mkurek (approved):

PR #528, opened 2016-10-11
Upload CSV KeyError bugfix

szok:


xor-xor (approved):

src/ralph_scrooge/rest/allocationclient.py
```diff
@@ -102,27 +102,32 @@ def get_allocation_from_file(file, allocation_admin=False):
...
-                row['environment']
-            )
-            errors.append(error)
+            if row.get('service_env_id'):
+                errors.append(
+                    'Service environment not found for ID: {}'.format(
+                        row.get('service_env_id')
+                    )
+                )
+            elif row.get('service_uid'):
```
mkurek: just else (ex. for case with 'service_name' instead of 'service_uid')

mkurek (changes requested):

mkurek (approved):

PR #530, opened 2016-10-14
Add endpoint for fetching usage values/costs per day/month.

xor-xor:
Docs will be added once we will have this new `v0.10` hierarchy finally discussed. By that time, this endpoint should be considered as private and subject to changes.

Please note that in `v0.10` we are dropping support for auth tokens in TastyPie format - from now on, only tokens in DjRF's format will be accepted (`v0.9` supports both of them - i.e. the same token can be provided in two ways).

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -111,8 +111,8 @@ Before uploading your usages, you need to do two things:
      ``/admin/ralph_scrooge/usagetype/add/``. Remember that every
      service usage type must have a unique symbol.
 
-    * leave "Show usage type in report" checked.
-    * select "Service usage type".
+    * leave "Show usage type in report" checked. (XXX such field doesn't exist?)
+    * In "Type" field select "Service usage type".
 
   2. Create Pricing Service by visiting
      ``/admin/ralph_scrooge/pricingservice/add/``. Possible options
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -111,8 +111,8 @@ Before uploading your usages, you need to do two things:
      ``/admin/ralph_scrooge/usagetype/add/``. Remember that every
      service usage type must have a unique symbol.
 
-    * leave "Show usage type in report" checked. (XXX such field doesn't exist?)
-    * In "Type" field select "Service usage type".
+    * leave "Active" box checked.
+    * In "Type" field, select "Service usage type".
 
   2. Create Pricing Service by visiting
      ``/admin/ralph_scrooge/pricingservice/add/``. Possible options
--- a/src/ralph_scrooge/rest/service_environment_costs.py
+++ b/src/ralph_scrooge/rest/service_environment_costs.py
@@ -398,7 +398,7 @@ def _create_trees(aggregated_costs):
     """
     cost_trees = {}
     for ac in aggregated_costs:
-        date_ = ac[0].date()
+        date_ = _get_truncated_date(ac[0])
         d = {
             ac[1]: {
                 '_type_symbol': ac[2],
--- a/src/ralph_scrooge/rest/service_environment_costs.py
+++ b/src/ralph_scrooge/rest/service_environment_costs.py
@@ -33,6 +33,15 @@ GROUP_BY_CHOICES = (
     ('month', 'month'),
 )
 
+
+# TODO(xor-xor): Think about some better name for this function and add some
+# caching to it.
+def get_parents():
+    return BaseUsage.objects.exclude(
+        pk__in=UsageType.objects.filter(usage_type='SU')
+    )
+
+
 class ServiceEnvironmentCostsDeserializer(Serializer):
     service_uid = serializers.CharField()
     environment = serializers.CharField()
@@ -42,15 +51,16 @@ class ServiceEnvironmentCostsDeserializer(Serializer):
     usage_types = ListField(serializers.CharField(), required=False)
 
     def validate_usage_types(self, attrs, source):
+        parents = get_parents()
         usage_types = attrs.get(source)
         if usage_types is None or len(usage_types) == 0:
-            attrs[source] = UsageType.objects.all()
+            attrs[source] = parents
             return attrs
         usage_types_validated = []
         unknown_values = []
         for ut in usage_types:
             try:
-                usage_types_validated.append(BaseUsage.objects.get(symbol=ut))
+                usage_types_validated.append(parents.get(symbol=ut))
             except BaseUsage.DoesNotExist:
                 unknown_values.append(ut)
         if len(unknown_values) > 0:
@@ -291,6 +301,36 @@ def _round_recursive(usages_and_costs):
             }
     return rounded
 
+def fetch_costs_alt(service_env, usage_types, date_from, date_to, group_by):
+
+    # aaa = DailyCost.objects_tree.filter(
+    #     date__gte=date_from,
+    #     date__lte=date_to,
+    #     service_environment=service_env,
+    # ).annotate(
+    #     cost_sum=Sum('cost'), value_sum=Sum('value')
+    # ).values_list('date', 'path', 'type__symbol', 'cost_sum', 'value_sum')
+
+    truncate_date = connection.ops.date_trunc_sql(group_by, 'date')
+    # aaa = DailyCost.objects_tree.filter(
+    #     date__gte=date_from,
+    #     date__lte=date_to,
+    #     service_environment=service_env,
+    # ).extra({group_by: truncate_date}).annotate(
+    #     cost_sum=Sum('cost'), value_sum=Sum('value')
+    # ).values_list('date', 'path', 'type__symbol', 'cost_sum', 'value_sum')
+
+    aaa = DailyCost.objects_tree.filter(
+        date__gte=date_from,
+        date__lte=date_to,
+        service_environment=service_env,
+    ).extra({group_by: truncate_date}).annotate(
+        cost_sum=Sum('cost'), value_sum=Sum('value')
+    )
+
+    from IPython import embed; embed(); assert False
+
+
 
 def fetch_costs(service_env, usage_types, date_from, date_to, group_by):
     """Fetch DailyCosts associated with given `service_env` and
@@ -404,13 +444,6 @@ def fetch_costs(service_env, usage_types, date_from, date_to, group_by):
         costs = _merge_costs_with_subcosts(
             results.get(date_, {}), results_subcosts.get(date_, {})
         )
-        for usage_type in usage_types:
-            if usage_type.symbol not in costs.keys():
-                costs.update({usage_type.symbol: {
-                    'cost': None,  # XXX(xor-xor) or maybe 0.0..?
-                    'usage_value': None,
-                    'subcosts': {},
-                }})
         costs_for_date = {
             'grouped_date': date_,
             'total_cost': round_safe(
@@ -444,7 +477,7 @@ class ServiceEnvironmentCosts(APIView):
             date_to = deserializer.object['date_to']
             group_by = deserializer.object['group_by']
 
-            costs = fetch_costs(
+            costs = fetch_costs_alt(
                 service_env, usage_types, date_from, date_to, group_by
             )
             if group_by == 'day':
--- a/src/ralph_scrooge/rest/service_environment_costs.py
+++ b/src/ralph_scrooge/rest/service_environment_costs.py
@@ -203,26 +203,28 @@ def _aggregate_costs(qs, usage_type, group_by, results):
     truncate_date = connection.ops.date_trunc_sql(group_by, 'date')
     qs_with_truncate_date = qs.extra({group_by: truncate_date})
     qs_by_usage_type = qs_with_truncate_date.filter(type=usage_type)
-    if qs_by_usage_type.exists():
+    # We use `len` instead of `exists` or `count` because of a bug in older
+    # Django, see: http://stackoverflow.com/q/37446551/5768173.
+    if len(qs_by_usage_type) > 0:
         # We assume that all elements in `qs_by_usage_type` have the same
         # `path` attribute - hence `qs_by_usage_type[0]`.
         path = qs_by_usage_type[0].path
-    results_by_date = qs_by_usage_type.values(group_by).annotate(
-        Sum('cost'), Sum('value')
-    ).order_by(group_by)
-    for r in results_by_date:
-        d = _get_truncated_date(r[group_by])
-        cost_and_usage_dict = {
-            path: {
-                '_usage_type_symbol': usage_type.symbol,
-                'cost': r['cost__sum'],
-                'usage_value': r['value__sum']
+        results_by_date = qs_by_usage_type.values(group_by).annotate(
+            Sum('cost'), Sum('value')
+        ).order_by(group_by)
+        for r in results_by_date:
+            d = _get_truncated_date(r[group_by])
+            cost_and_usage_dict = {
+                path: {
+                    '_usage_type_symbol': usage_type.symbol,
+                    'cost': r['cost__sum'],
+                    'usage_value': r['value__sum']
+                }
             }
-        }
-        if results.get(d) is not None:
-            results[d].update(cost_and_usage_dict)
-        else:
-            results[d] = cost_and_usage_dict
+            if results.get(d) is not None:
+                results[d].update(cost_and_usage_dict)
+            else:
+                results[d] = cost_and_usage_dict
 
 
 def _merge_costs_with_subcosts(costs, subcosts):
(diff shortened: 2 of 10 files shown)


doc/api.rst
```diff
@@ -111,8 +111,8 @@ Before uploading your usages, you need to do two things:
      ``/admin/ralph_scrooge/usagetype/add/``. Remember that every
      service usage type must have a unique symbol.
 
-    * leave "Show usage type in report" checked.
-    * select "Service usage type".
+    * leave "Show usage type in report" checked. (XXX such field doesn't exist?)
```
mkurek: afair it's `active` right now

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,455 @@
...
+USAGE_COST_NUM_DIGITS = 2
+USAGE_VALUE_NUM_DIGITS = 5
+
+
+class ServiceEnvironmentCostsDeserializer(Serializer):
+    service_uid = serializers.CharField()
+    environment = serializers.CharField()
+    date_from = serializers.DateField()
+    date_to = serializers.DateField()
+    group_by = serializers.CharField(default='day')
```
mkurek: doesn't DRF have ChoiceField? http://www.tomchristie.com/rest-framework-2-docs/api-guide/fields#choicefield

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,455 @@
...
+                    ['"{}"'.format(v) for v in allowed_values]
+                ))
+            )
+            raise serializers.ValidationError(err)
+        return attrs
+
+    def validate_usage_types(self, attrs, source):
+        usage_types = attrs.get(source)
+        if usage_types is None or len(usage_types) == 0:
+            attrs[source] = UsageType.objects.all()
```
mkurek: here you return all `UsageType`s, but later you check for `BaseUsage` - what case is valid?

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,455 @@
...
+
+class ServiceEnvironmentDailyCostsSerializer(Serializer):
+    service_environment_costs = DailyCostsSerializer(many=True)
+
+
+class ServiceEnvironmentMonthlyCostsSerializer(Serializer):
+    service_environment_costs = MonthlyCostsSerializer(many=True)
+
+
+def date_range(start, stop, step=timedelta(days=1)):
```
mkurek: you could use `dateutil.rrange` as well

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,455 @@
...
+    truncate_date = connection.ops.date_trunc_sql(group_by, 'date')
+    initial_qs = DailyCost.objects_tree.filter(
+        service_environment=service_env,
+        date__gte=date_from,
+        date__lte=date_to,
+    )
+    qs = initial_qs.extra({group_by: truncate_date})
+    total_costs = _get_total_costs(qs, group_by)
+    results = {}
+    for usage_type in usage_types:
```
mkurek: please annotate by path - then you'll have to do just one PR - and you could post-process it in python (it should be faster than querying multiple times this table - afaik path field doesn't have index as well). 

The "flow" here could be as follows:
- filter daily costs by service-env and date range
- group queryset by date (month?) and path
- do a postprocessing in python - create some data structure for storing costs by date, and nested one, where you'll store costs by following parts of `path` (splitted by `/`)

The result should be exactly the same, but you'll make much less DB queries (1 instead of N)

You could also consider calculating total cost in python (you have to iterate over each "row" of result either way, so summing total cost up shouldn't be a problem here, but calling to DB might be) :)

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,455 @@
...
+                'cost': round_safe(v['cost'], USAGE_COST_NUM_DIGITS),
+                'usage_value': round_safe(
+                    v['usage_value'], USAGE_VALUE_NUM_DIGITS
+                ),
+                'subcosts': _round_recursive(v.get('subcosts', {})),
+            }
+    return rounded
+
+
+def fetch_costs(service_env, usage_types, date_from, date_to, group_by):
```
mkurek: `usage_types` might be misleading here - it is (or should be) list of `BaseUsage` instances, not `UsageType`

mkurek (approved):

mkurek (changes requested):
@xor-xor what do you think about this groupping?

doc/api.rst

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Done. Although it leads me to define choices as:

```
GROUP_BY_CHOICES = (
    ('day', 'day'),
    ('month', 'month'),
)
```

...which IMO looks kind of weird - but on the other hand, we get validation of this field for free, which is a good thing.

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Both of them..? :wink: 

I mean, this later check for `BaseUsage`s if for the case when someone would provide a `PricingService` in `usage_type` field.

So maybe I should return all `BaseUsage`s here (instead of `UsageType`s)..? But to be honest - this relation between `BaseUsage` and `UsageType` and `PricingService` is quite confusing to me (i.e. that they are all kind of equal in this context - especially `PricingService` being a subclass of `BaseUsage`) :wink:

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Yes, it really is misleading/confusing (see one of my previous comments).

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: You mean `dateutil.rrule`..? I'm not sure if it's really needed here - for instance, it returns `datetime.datetime` objects, while this module uses `datetime.date` exclusively, so we would have to convert its results. And "my" `date_range` is only 3 lines long, so... :wink:

mkurek: sure - it could stay as it is

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Regarding the first part of your comment - I'll try to provide an alternative version of `fetch_costs` function, so we'd be able to compare them performance-wise.

And speaking of the second part (calculation of total cost) - but what about the fact that total cost is not the sum of costs calculated for selected usage types..? IMO we have to make a separate DB query here anyway.

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Done.

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,636 @@
...
+            )
+        except ServiceEnvironment.DoesNotExist:
+            errors.append(
+                'service environment for service with UID "{}" and '
+                'environment "{}" does not exist'.format(service_uid, env)
+            )
+        else:
+            attrs['_service_environment'] = service_env
+
+        if len(errors) > 0:
```
mkurek: maybe just `if errors`? :)

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,636 @@
...
+    grouped by date period given as `group_by` (e.g., "day", "month").
+    `results` argument is a dictionary, where results of this function are
+    appended (i.e., this function doesn't return anything).
+    """
+    # XXX(xor-xor): Check if truncate_date is really needed here.
+    truncate_date = connection.ops.date_trunc_sql(group_by, 'date')
+    qs_with_truncate_date = qs.extra({group_by: truncate_date})
+    qs_by_type = qs_with_truncate_date.filter(type=type_)
+    # We use `len` instead of `exists` or `count` because of a bug in older
+    # Django, see: http://stackoverflow.com/q/37446551/5768173.
```
mkurek: we have Django 1.10 now :)

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,636 @@
...
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAuthenticated, IsServiceOwner)
+
+    def post(self, request, *args, **kwargs):
+        deserializer = ServiceEnvironmentCostsDeserializer(data=request.data)
+
+        # A workaround for ListField validation (from drf_compound_fields) that
+        # for some reason can't handle nulls gracefully (it gives TypeError
+        # instead of ValidationError or something like that).
+        # This workaround can be removed once we switch to DjRF 3.x and get rid
```
mkurek: we have drf 3.x now - you could remove it :)

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,636 @@
...
+    """
+    # TODO(xor-xor): Remember to remove `fetch_costs` or `fetch_costs_alt`
+    # function (with dependencies) once we have results of the aforementioned
+    # performance tests.
+    truncate_date = connection.ops.date_trunc_sql(group_by, 'date')
+    initial_qs = DailyCost.objects_tree.filter(
+        date__gte=date_from,
+        date__lte=date_to,
+        service_environment=service_env,
+    ).extra({group_by: truncate_date})
```
mkurek: `extra` is deprecated now (https://docs.djangoproject.com/en/1.10/ref/models/querysets/#django.db.models.query.QuerySet.extra) - we could use `TruncMonth` here IMO (https://docs.djangoproject.com/en/1.10/ref/models/database-functions/#django.db.models.functions.datetime.TruncMonth or http://stackoverflow.com/a/8746532/761782)

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,636 @@
...
+    cost_trees_filtered = _filter_by_types(cost_trees_, types)
+
+    final_result = {
+        'service_environment_costs': []
+    }
+    if group_by == 'month':
+        delta = relativedelta(months=1)
+        date_range_ = date_range(
+            date_from.replace(day=1),
+            date_to.replace(day=1) + delta,
```
mkurek, line 366: maybe `date_range` should iterate to `date_to` including it (`<=` instead of `<`)? Then we wouldn't have to add this `+delta` here and when we group by day (to increase it by one)?

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,636 @@
...
+    truncate_date = connection.ops.date_trunc_sql(group_by, 'date')
+    initial_qs = DailyCost.objects_tree.filter(
+        date__gte=date_from,
+        date__lte=date_to,
+        service_environment=service_env,
+    ).extra({group_by: truncate_date})
+
+    total_costs = _get_total_costs(initial_qs, group_by)
+
+    aggregated_costs = initial_qs.values_list(
```
mkurek: wouldn't it be better to use `values` here instead of `values_list`? Then we'd get rid of this "magic" indexes like `ac[5]` and use field names like `ac['path']`

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,636 @@
...
+def fetch_costs_alt(service_env, types, date_from, date_to, group_by):
+    """An alternative version of `fetch_costs` function. The difference is that
+    it makes less DB queries, so in theory it should be faster (to be confirmed
+    with performance tests).
+    """
+    # TODO(xor-xor): Remember to remove `fetch_costs` or `fetch_costs_alt`
+    # function (with dependencies) once we have results of the aforementioned
+    # performance tests.
+    truncate_date = connection.ops.date_trunc_sql(group_by, 'date')
+    initial_qs = DailyCost.objects_tree.filter(
```
mkurek, line 326: we should filter here `depth` to be max 2

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,636 @@
...
+                }
+        else:
+            tmp_dict[k] = v
+    cost_trees_.update(tmp_dict)
+    return cost_trees_
+
+
+def _filter_by_types(cost_trees, types):
+    """Makes a copy of `cost_trees` and filters out top-level costs from it
+    if their types are not in `types`.
```
mkurek, line 495: please describe here why we can't do this at database level

src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -0,0 +1,636 @@
...
+
+    The pairing between costs and subcosts is done via the path component (e.g.
+    a cost with '484/483' is a subcost of '484').
+    """
+    cost_trees = {}
+    for ac in aggregated_costs:
+        date_ = _get_truncated_date(ac[0])
+        d = {
+            ac[1]: {
+                '_type_symbol': ac[2],
```
mkurek: add some comment here why couldn't we use type symbol directly as a key in `d` dict

mkurek (changes requested):

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: I know, this is exactly what I'm going to do (I'm in the middle of a transition to Django 1.10 / DjRF 3.x) :wink:

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: I know - I'm in the middle of a transition to Django 1.10 / DjRF 3.x.

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Hmm... As far as I remember, we both decided that the maximum level of nesting in DailyCosts will be 1, so where that new value comes from..?

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: I know :wink: As I stated in my previous comments - I'm in the middle of a transition to this version.

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Done.

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Done.

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Done.

src/ralph_scrooge/rest/service_environment_costs.py
mkurek: It is possible that there will be costs with greater depth. By default it's turned off (using `SAVE_ONLY_FIRST_DEPTH_COSTS` in settings), but when you turn in on (setting it to `False`), for each `PricingService` with `universal` "calculating model" (not only the ones with fixes prices) whole subtree of costs will be saved (for example, PricingService 1 is dependent on PricingService2 which has costs from some base UsageTypes, Extra Costs and Teams - theoretically it could be nested infinitely). 

Cutting of here at `depth__lte=1` (or `depth__in=(0,1)` - I don't know which will be faster for DB) will work in cases when someone has this variable (`SAVE_ONLY_FIRST_DEPTH_COSTS` set to `False`).

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Done.

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Done (I went with `depth__lte=1`, as it seems to be more idiomatic than the latter form).

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: I'm over the fence with this idea :wink: On the one hand - you're right - it would simplify the code a little bit - but on the other - it would introduce some inconsistency / unexpected behaviour related to how "normal" `range` behaves in Python. Personally, I would be rather surprised to see that any iterator having `_range` in its name would include the last value of the interval given as argument.

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: You mean why it starts with `_`..? That's because it's just a temporary key, that is removed later in the process, so anyone dealing with this code shouldn't rely on it. But yes, you're right with adding a comment.

xor-xor: Done.

src/ralph_scrooge/rest/service_environment_costs.py
mkurek: good point with python range. So leave it as it is :)

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Done (I've added a comment related to this, but in a different, more suitable location).

src/ralph_scrooge/rest/service_environment_costs.py
xor-xor: Good point (done).

mkurek (approved):

szok (approved):

PR #532, opened 2016-10-25
Add mgmt command for (re)calculation of daily costs per single day.

xor-xor:


src/ralph_scrooge/management/commands/scrooge_calculate_dailycosts.py
```diff
@@ -0,0 +1,156 @@
...
+        return CostDateStatus.objects.filter(
+            date=date_,
+            **{'forecast_calculated' if forecast else 'calculated': True}
+        ).values_list('date', flat=True).exists()
+
+    def _has_accepted_costs(self, date_, forecast):
+        return CostDateStatus.objects.filter(
+            date=date_,
+            **{'forecast_accepted' if forecast else 'accepted': True}
+        ).values_list('date', flat=True).exists()
```
mkurek: no need for `values_list` to check if any record `exists`

src/ralph_scrooge/management/commands/scrooge_calculate_dailycosts.py
```diff
@@ -0,0 +1,156 @@
...
+            )
+            return
+
+        # Determine which plugins are needed and finally calculate & save the
+        # costs.
+        plugins = []
+        for p in collector.get_plugins():
+            if p.name in pricing_service_names_verified:
+                plugins.append(p)
+        processed_costs = collector.process(date_, forecast, plugins=plugins)
```
mkurek: we could think about wrapping these 3 calls to collector into some Collector's method, which will have date (start and end?), forecast and plugins param and will do the rest - @xor-xor what do you think?

mkurek (changes requested):

src/ralph_scrooge/management/commands/scrooge_calculate_dailycosts.py
xor-xor: Ah, right - I'm a copy & paste victim :wink:

src/ralph_scrooge/management/commands/scrooge_calculate_dailycosts.py
xor-xor: Makes sense, but only if this method will be used anywhere else apart from this mgmt command - do you see any use for it in the future..?

On the other hand, I see that `Collector` already has methods like `process_period` and`process`, so adding another one, with similar name/behavior can be confusing to ppl working with it later.

xor-xor: Done.

mkurek (approved):

src/ralph_scrooge/management/commands/scrooge_calculate_dailycosts.py
```diff
@@ -0,0 +1,149 @@
...
+        query_params = {
+            'active': True,
+            'plugin_type': PricingServicePlugin.pricing_service_fixed_price_plugin.id,  # noqa: E501
+        }
+        # ...but when `-p` option is present, take only PricingServices
+        # specified with it.
+        if len(pricing_service_names) > 0:
+            query_params.update({'name__in': pricing_service_names})
+        pricing_services = PricingService.objects.filter(**query_params)
+        pricing_service_names_verified = [ps.name for ps in pricing_services]
```
szok: Why not
pricing_services.vlaues_list('name', flat=True)
?

szok (approved):

PR #544, opened 2016-11-08
Notifications about anomalies in usages

xor-xor:
Detect missing values and unusually big changes in uploaded usage types and notify their owners.

This PR adds management command which should be launched on daily basis (e.g. from cron). It analyzes usage types uploaded during the month-wide window, which ends on the date give via `-e` argument.

It looks for two types of anomalies: missing usage types for given day, and unusually big changes of values between two consecutive days, where by "unusually big" it understands a relative change calculated with `(d1 - d2) / d2`, compared to the value stored in `UsageType.change_tolerance` (`0.2` by default). In case of detecting any of such anomalies, all the owners of a given usage type will be notified by e-mail.

It also implements a "tolerance margin" for missing values, accounting for the cases where usage types are not expected due to longer upload frequency (e.g. monthly instead of daily), and/or where the owners shouldn't receive e-mails because they won't be able to react anyway (e.g. during the weekends).


src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
+# -*- coding: utf-8 -*-
```
mkurek: don't name this command like `scrooge_XXX` - this prefix was useful when scrooge was tied to ralph. Now, when scrooge is standalone, this prefix is not necessary anymore.

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import datetime
+import logging
+from dateutil.relativedelta import relativedelta
+
+from collections import OrderedDict
```
mkurek: collections (unlike dateutil) is in stdlib - change order of imports

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+from django.template.loader import render_to_string
+
+from ralph_scrooge.models import UsageType, DailyUsage
+# TODO(xor-xor): Consider moving `date_range` to some utils module.
+from ralph_scrooge.rest.service_environment_costs import date_range
+
+
+log = logging.getLogger(__name__)
+
+# XXX This could be overridden on per-UsageType basis (e.g. in some dict loaded
```
mkurek: 👍

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+
+def get_usage_types(names):
+    if names:
+        ut = UsageType.objects.filter(name__in=names).order_by('name')
+    else:
+        ut = UsageType.objects.order_by('name')
+    return ut
+
+
+def get_negative_month_range(end_date=None):
```
mkurek: why negative?

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+        date=date, type=usage
+    ).values('date').aggregate(Sum('value'))['value__sum']
+
+
+def _get_usage_values_for_date_range(usage_types):
+    results = OrderedDict()
+    for usage in usage_types:
+        results[usage] = OrderedDict()
+        for date_ in get_negative_month_range():
+            results[usage][date_] = _get_usage_value(
```
mkurek: fetch all usages at once. Sth like:

```
daily_usages = dict(DailyUsage.objects.filter(date__gte=start_date, date__lte=end_date, type=usage).values_list('date').annotate(Sum('value'))
for date_ in get_negative_month_range():
   results[usage][date_] = daily_usages.get(date_, 0)
```

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+    return date_range(start_date, end_date)
+
+
+def _get_usage_value(date, usage):
+    return DailyUsage.objects.filter(
+        date=date, type=usage
+    ).values('date').aggregate(Sum('value'))['value__sum']
+
+
+def _get_usage_values_for_date_range(usage_types):
```
mkurek: shouldn't this method take `end_date` as a param?

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+        results[usage] = OrderedDict()
+        for date_ in get_negative_month_range():
+            results[usage][date_] = _get_usage_value(
+                date_, usage
+            )
+    return results
+
+
+def _detect_missing_values(usage_values):
+    missing_values = OrderedDict()
```
mkurek: does it have to be ordered dict? maybe you could use defaultdict?

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+        # user:
+        # .first_name, .last_name, .username, .email
+
+
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '-u',
+            dest='usage_names',  # XXX what about using symbols here..?
```
mkurek: 👍 for symbols

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+        anomalies = _merge_and_group_by_usage_type(
+            usage_types, missing_values, big_changes_by_type
+        )
+        anomalies_to_report = _group_anomalies_by_owner(anomalies)
+        if not dry_run:
+            _send_notifications(anomalies_to_report)
+
+
+def pprint_anomalies(anomalies):
+    """XXX This is just a temporary function for debugging/development."""
```
mkurek: i'll keep it (maybe improve it a little bit) to get some output during dry-run

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+            continue
+        merged_anomalies[usage] = {
+            'big_changes': big_changes,
+            'missing_values': missing_values_,
+        }
+    return merged_anomalies
+
+
+def _group_anomalies_by_owner(anomalies):
+    ret = OrderedDict()
```
mkurek: defaultdict?

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
```diff
@@ -0,0 +1,274 @@
...
+
+def _merge_and_group_by_usage_type(usage_types, missing_values, big_changes_by_type):  # noqa: E501
+    merged_anomalies = OrderedDict()
+    for usage in usage_types:
+        big_changes = big_changes_by_type.get(usage)
+        missing_values_ = missing_values.get(usage)
+        if big_changes is None and missing_values is None:
+            continue
+        merged_anomalies[usage] = {
+            'big_changes': big_changes,
```
mkurek: we should figure out some better name for this :)

src/ralph_scrooge/models/usage.py
```diff
@@ -64,6 +73,26 @@ class UsageType(BaseUsage):
         related_name='excluded_usage_types',
         blank=True,
     )
+    owners = db.ManyToManyField(
+        'ScroogeUser',
+        verbose_name=_("Owners of usage type"),
+        related_name='usage_type_owners',
+        blank=True,
+    )
+    upload_freq = db.PositiveIntegerField(
```
mkurek, line 97: this param should be (somehow) considered in anomalies detection. in my opinion we should take some margin (ex. 2 days - it could be general for all frequencies or different for each) and expect uploads for day/week/month after end of the period (end of day, end of week, and of month + some margin) for this period of time. Example (considering margin as 2 days):
- for daily frequency: on 8.11 we should expect uploads for 6.11 and before
- for weekly frequency: on 8.11 (end of 31.10 - 6.10 week + 2 days) we should expect uploads for week 31.10 - 6.11 (but, on 9.11, 10.11, ..., 14.11 we should expect usages upload for max. 6.11, not later)
- for monthly frequency: on 2.11 (end of october + 2 days margin) we should expect uploads for whole october (but on 3.11, ..., 1.12 we should expect uploads for max. 31.10, not later)

mkurek (changes requested):

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: Actually, I have a better idea - instead of putting this into settings, I'd add the `change_tolerance` (or `diff_tolerance`) field to `UsageType`, with default value `0.2` - that will allow to customize it by owners of given usage type via admin panel (and I guess that this default value won't be OK for everybody, so...).

What do you think about this..?

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: Because we get iterator over range, where `end_date` is yesterday (by default), and `start_date` is yesterday - 1 month.
But if you have some better idea for the name here, I'm open for suggestions :wink:

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: Yes, sure - it's just a temporary one. Let me know if you have some ideas for this.

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: I needed OrderedDict for development purposes only, so now almost all occurences of it are replaced with normal dicts (except for the place where I prepare the list of dates for e-mail template - I need them to be sorted there).

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: "Normal" one is just enough here.

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: Yup (done).

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: What about `unusual_changes` or `spikes`..?

src/ralph_scrooge/management/commands/scrooge_detect_usage_anomalies.py
xor-xor: Right (done).

src/ralph_scrooge/models/usage.py
xor-xor: Done.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -0,0 +1,442 @@
...
+
+
+class UnknownUsageTypeUploadFreqError(Exception):
+    """Raised when UsageTypeUploadFreq contains a choice which is not handled
+    in this module (see: `_get_max_expected_date` function).
+    """
+    pass
+
+
+log = logging.getLogger(__name__)
```
mkurek: we use `logger` everywhere for such things- could you change it to keep it compatible with the rest of the code?

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -0,0 +1,442 @@
...
+    freq_margin = UsageTypeUploadFreq.from_id(usage.upload_freq).margin
+    if freq_name == 'daily':
+        max_date = date - freq_margin
+    elif freq_name == 'weekly':
+        max_date = date - freq_margin + relativedelta(weekday=rd.SU(-1))
+    elif freq_name == 'monthly':
+        if date.day == freq_margin.days:
+            max_date = date - freq_margin
+        else:
+            max_date = date - freq_margin + relativedelta(day=1, days=-1)
```
mkurek, line 137: could you describe it briefly? as far as i understand this is going to return last day of previous month?

src/ralph_scrooge/migrations/0003_auto_20161117_1248.py
```diff
@@ -0,0 +1,36 @@
...
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ralph_scrooge', '0002_auto_20161104_1323'),
```
mkurek: probably https://github.com/allegro/ralph_pricing/pull/551 will be merged faster, so remember to bump migration number and dependent number

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -0,0 +1,442 @@
...
+        daily_usages = dict(
+            DailyUsage.objects.filter(
+                date__gte=start_date, date__lte=end_date, type=usage
+            ).values_list('date').annotate(Sum('value'))
+        )
+        results[usage].update(daily_usages)
+    return results
+
+
+def _get_max_expected_date(usage, date):
```
mkurek: could you write some unit tests for this method?

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -0,0 +1,442 @@
...
+            log.warning(
+                "Unknown symbol(s) detected: {}. Ignoring."
+                .format(", ".join(unknown_symbols))
+            )
+    else:
+        uts = UsageType.objects.order_by('symbol')
+    return uts
+
+
+def get_negative_month_range(end_date):
```
mkurek: Is it possible to make this method more generic - it could take number of months or days (to subtract) as a param?

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
mkurek: And there is a "bug" IMO - when "monthly" freq is used, couple of first days of the month will never be verified. For example:
assuming today is 2016.11.02, we're assuming, that usages for October might not be uploaded - so `max_date` is 2016.09.30 (so no day will be verified, because `min_date` (`start_date`) is `2016.11.02 - month == 2016.10.02`). Assuming today is 2016.11.03, usages for October should be uploaded, so `max_date` is 2016.10.31 - but, again, start date is `2016.11.03 - month == 2016.10.03` - what about 2016.10.01 and 2016.10.02? They will be not verified at all.

Possible solution for that might be to set start_date for some more than month ago (ex. 40 days) - but the perfect here would be to parametrize it somewhere.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -0,0 +1,442 @@
...
+    final e-mails with anomaly reports.
+    """
+    usage_values = _get_usage_values_for_month(usage_types, end_date)
+    missing_values = _detect_missing_values(usage_values, end_date)
+    unusual_changes = _detect_unusual_changes(usage_values, end_date)
+    if not missing_values and not unusual_changes:
+        return
+    # Post-processing.
+    anomalies = _merge_by_type(usage_types, missing_values, unusual_changes)
+    anomalies_to_report = _group_anomalies_by_owner(anomalies)
```
mkurek: could we move grouping by owner out of this function? `_detect_anomalies` will probably be reused in GUI someday, but we'll have to present anomalies by type only. Grouping by owner is necessary only for mailing.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -0,0 +1,442 @@
...
+        if not owners.exists():
+            log.warning(
+                'Anomalies detected in UsageType "{}", but it doesn\'t have '
+                'any owners defined, hence we cannot notify them. Please '
+                'correct that ASAP.'.format(usage.name)
+            )
+            continue
+        for owner in owners:
+            if ret.get(owner) is None:
+                ret[owner] = {'missing_values': {}, 'unusual_changes': {}}
```
mkurek: consider defaultdicts here

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -0,0 +1,442 @@
...
+
+    def group_missing_values_by_date(usage, missing_values):
+        ret = {}
+        for date in missing_values:
+            if ret.get(date) is None:
+                ret[date] = set()
+            ret[date].add(usage)
+        return ret
+
+    ret = {}
```
mkurek: consider defaultdicts here

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -0,0 +1,442 @@
...
+    """A wrapper around Django's own `send_mail` with error-checking and
+    logging added.
+    """
+    try:
+        num_msgs_sent = send_mail(
+            (
+                'Unusual changes / missing values for your usage type(s) in '
+                'Scrooge'
+            ),
+            '',  # TODO(xor-xor): Do we need plain-text version of the message?
```
mkurek: no

mkurek (changes requested):

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: :+1:

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -0,0 +1,442 @@
...
+            'site_url': settings.BASE_MAIL_URL,
+        }
+        body = render_to_string(template_name, context)
+        _send_mail(recipient.email, body)
+
+
+def pprint_anomalies(anomalies):
+    """Pretty-printing for `--dry-run` switch and for debugging."""
+    for user, anomalies_ in anomalies.items():
+        print("-" * 79)
```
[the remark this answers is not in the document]
xor-xor: Done.

src/ralph_scrooge/migrations/0003_auto_20161117_1248.py
xor-xor: OK.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Done.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Done & done.
Actually, the solution for the "bug" above was quite simple - I've added a param `align_to_month` to `get_negative_month_range`, which is set to `True` when `UsageType.upload_freq` is set to `'monthly'` - it extends the returned range to the beginning of month.

src/ralph_scrooge/models/usage.py
```diff
@@ -64,6 +88,40 @@ class UsageType(BaseUsage):
...
+        null=False,
+        blank=False,
+        default=1,
```
szok: Better use:
default=UsageTypeUploadFreq. daily.id

src/ralph_scrooge/templatetags/joinby.py
```diff
@@ -0,0 +1,13 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def joinby(values, sep):
```
szok, line 7: 👍

szok (approved):

src/ralph_scrooge/models/usage.py
xor-xor: Done.

mkurek (approved):

PR #546, opened 2016-11-10
New param in service-environment-costs endpoint: forecast

mkurek:
Set it to true to fetch forecast costs (by default "regular" costs are returned)

### DIFF
--- a/src/ralph_scrooge/rest/service_environment_costs.py
+++ b/src/ralph_scrooge/rest/service_environment_costs.py
@@ -36,7 +36,7 @@ GROUP_BY_CHOICES = (
 )
 
 
-@memoize
+# @memoize
 def get_valid_types():
     """We are interested here in all types except "SU" which is "Service usage
     type" (see TYPE_CHOICES in ralph_scrooge.models.usage).
--- a/src/ralph_scrooge/rest/service_environment_costs.py
+++ b/src/ralph_scrooge/rest/service_environment_costs.py
@@ -58,6 +58,7 @@ class ServiceEnvironmentCostsDeserializer(Serializer):
     types = serializers.ListField(
         child=serializers.CharField(), required=False
     )
+    forecast = serializers.BooleanField(default=False, required=False)
 
     def validate_types(self, types):
         valid_types = get_valid_types()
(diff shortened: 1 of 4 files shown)


src/ralph_scrooge/rest/service_environment_costs.py
```diff
@@ -36,7 +36,7 @@
 )
 
 
-@memoize
+# @memoize
```
xor-xor: Why is that commented out..?

src/ralph_scrooge/tests/rest/test_service_environment_costs.py
```diff
@@ -284,6 +286,37 @@ def test_if_grouped_date_has_correct_format_when_group_by_month_given(self):  #
         date_received = json.loads(resp.content)['service_environment_costs'][0]['grouped_date']  # noqa: E501
         self.assertEquals(date_received, date_expected)
 
+    def test_for_forecast_costs(self):  # noqa: E501
```
xor-xor: `noqa` is not needed here, I guess..?

src/ralph_scrooge/utils/common.py
```diff
@@ -6,10 +6,10 @@
 from __future__ import unicode_literals
 
 from decimal import Decimal
-from functools import wraps
 
 from django.conf import settings
-from ralph_scrooge.utils.cache import memoize as memoize_orig
+# DEPRECATED. Import memoize directly from `ralph_scrooge.cache`
+from ralph_scrooge.utils.cache import memoize  # noqa: F401
```
xor-xor, line 12: So... why do we need this import here..?

xor-xor (changes requested):

src/ralph_scrooge/rest/service_environment_costs.py
mkurek: right - I should revert it

src/ralph_scrooge/utils/common.py
mkurek: because still lot of modules import `memoize` from there

xor-xor: OK.

xor-xor (approved):

szok (approved):

PR #551, opened 2016-11-17
Add IPInfo model and migrate data from PricingObject to IPInfo

szok:


src/ralph_scrooge/migrations/0004_copy_ipinfo.py
```diff
@@ -0,0 +1,60 @@
...
+        pricing_objects = PricingObject.objects.filter(
+            name=d['name'], type=PRICING_OBJECT_TYPES.IP_ADDRESS
+        ).order_by('created')
+        # Default PO for the oldest entry
+        default_pricing_object = pricing_objects[0]
+        for po in pricing_objects[1:]:
+            DailyCost.objects.filter(pricing_object=po).update(
+                pricing_object=default_pricing_object
+            )
+            DailyPricingObject.objects.filter(pricing_object=po).update(
```
mkurek: use here DailyUsage rather than DailyPricingObject. Create daily pricing object for today (`dpo = default_pricing_object.get_daily_pricing_object(today)`) and run sth like `DailyUsage.objects.filter(daily_pricing_object__pricing_object=po).update(daily_pricing_object=dpo)`

mkurek (changes requested):

src/ralph_scrooge/migrations/0004_copy_ipinfo.py
```diff
@@ -0,0 +1,73 @@
...
+    # Find duplicates
+    duplicates = PricingObject.objects.filter(
+        type=PRICING_OBJECT_TYPES.IP_ADDRESS
+    ).values(
+        'name'
+    ).annotate(Count('name')).order_by().filter(
+        name__count__gt=1,
+    )
+
+    for d in duplicates:
```
mkurek: add some log/print here what ip is currently processed

mkurek (changes requested):

mkurek (approved):

src/ralph_scrooge/models/pricing_object.py
```diff
@@ -175,6 +176,26 @@ def save(self, *args, **kwargs):
...
+        max_digits=39,
+        decimal_places=0,
+        default=None,
+    )
+
+    class Meta(PricingObject.Meta):
+        app_label = 'ralph_scrooge'
+
+    def save(self, *args, **kwargs):
+        self.number = int(ipaddress.ip_address(self.name or 0))
```
xor-xor: Can we safely assume, that we will never get a `ValueError` from `ip_address` here..? Or putting it in other way: does `self.name` guarantee to always hold a valid IPv4/IPv6 address..?

mkurek:
@szok bump migrations numbers

src/ralph_scrooge/models/pricing_object.py
```diff
@@ -175,6 +177,29 @@ def save(self, *args, **kwargs):
...
+    )
+
+    class Meta(PricingObject.Meta):
+        app_label = 'ralph_scrooge'
+
+    def save(self, *args, **kwargs):
+        try:
+            self.number = int(ipaddress.ip_address(self.name))
+        except (ipaddress.AddressValueError, ValueError):
+            raise ValidationError(_('Is not a valid ip address'))
```
xor-xor: s/ip/IP/ :wink:

xor-xor (approved):

PR #557, opened 2016-11-21
Fetch cloud projects from all providers from Ralph3

mkurek:


xor-xor (approved):
Fix this failing test and it will be OK for me.

PR #558, opened 2016-11-24
Add filtering by service on `pricingserviceusage` endpoint in API.

quamilek:
In some reason user would like to filter stored usages coming from
one of service, not fetching usages from all services. This pull
request implement filtering `pricingserviceusages` endpoint by
`service_id` parameter in GET method.

### DIFF
--- a/src/ralph_scrooge/rest/pricing_service_usages.py
+++ b/src/ralph_scrooge/rest/pricing_service_usages.py
@@ -9,6 +9,7 @@ import logging
 from collections import defaultdict
 from datetime import datetime
 
+from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
 from django.http import HttpResponse
 from rest_framework import status
@@ -27,6 +28,7 @@ from ralph_scrooge.models import (
     DailyUsage,
     PricingObject,
     PricingService,
+    Service,
     ServiceEnvironment,
     UsageType,
 )
@@ -298,6 +300,7 @@ def list_pricing_service_usages(
         request, usages_date, pricing_service_id, *args, **kwargs
 ):
     err_msg = None
+    service = None
     try:
         pricing_service = PricingService.objects.get(id=pricing_service_id)
     except PricingService.DoesNotExist:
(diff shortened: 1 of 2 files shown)


mkurek (approved):

src/ralph_scrooge/tests/rest/test_pricing_service_usages.py
```diff
@@ -1034,19 +1034,30 @@ def test_posted_pricing_service_usage_is_the_same_when_fetched_with_get(self):
...
+            }
         )
+        resp = self.client.get(url)
         self.assertEquals(resp.status_code, 200)
         received_response = json.loads(resp.content)
+        usages = received_response['usages']
+        self.assertEqual(len(usages), 2)
         self.assertNestedDictsEqual(expected_response, received_response)
 
+        # check filtering by `service_id`
```
xor-xor: I'd rather create a separate test for this, instead of adding it to another, which is quite lengthy by itself anyway.

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -319,11 +332,11 @@ def list_pricing_service_usages(
     if err_msg is not None:
         return Response({'error': err_msg}, status=status.HTTP_400_BAD_REQUEST)
 
-    usages = get_usages(usages_date, pricing_service)
+    usages = get_usages(usages_date, pricing_service, service)
     return Response(PricingServiceUsageSerializer(usages).data)
 
 
-def get_usages(usages_date, pricing_service):
+def get_usages(usages_date, pricing_service, service=None):
```
xor-xor: What about renaming `service` to `filter_by_service` here..?

xor-xor: Please update docstring for this function.

src/ralph_scrooge/rest/pricing_service_usages.py
```diff
@@ -306,6 +309,16 @@ def list_pricing_service_usages(
             .format(pricing_service_id)
         )
 
+    service_id = request.GET.get('service_id')
+    if service_id:
+        try:
+            service = Service.objects.get(id=service_id)
+        except ObjectDoesNotExist:
+            err_msg = (
```
xor-xor: No need to split this into 4 lines, 1 should be enough.

xor-xor (changes requested):

xor-xor (approved):

PR #563, opened 2016-11-28
[API] Add static schema and swagger-ui

xor-xor:
This PR also adds a mgmt command for generation of API schema in JSON format, and bumps DjRF's version (from 3.5.0 to 3.5.3).

Please note, that `host` field in schema (YAML) still has to be filled in manually.


src/ralph_scrooge/rest/swagger.py
```diff
@@ -0,0 +1,15 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from django.shortcuts import render_to_response
+from django.views.generic import View
+
+
+class BootstrapSwagger(View):
```
mkurek, line 12: why Bootstrap? Isn't it copied from one of already existing views? :)

mkurek (approved):

src/ralph_scrooge/urls.py
```diff
@@ -66,6 +68,11 @@
         ServiceEnvironmentCosts.as_view(),
         name='service_environment_costs',
     ),
+    url(
+        r'^scrooge/api/$',
```
mkurek, line 73: consider linking it to `/api/` too

src/ralph_scrooge/rest/swagger.py
xor-xor: Because this bootstrapping view does just that, i.e. it just serves static html from which `bootstrap-ui` (js) is launched.

Or maybe I've misunderstood your question..? :wink:

src/ralph_scrooge/urls.py
xor-xor: I'll add a 302 redirect from `/api/` to `/scrooge/api/` - that's because [here](https://github.com/xor-xor/ralph_pricing/blob/3e4d31653b2c674a50d2b4b72f606dd8c2bd5bdc/src/ralph_scrooge/media/api_schema.yaml#L18) we can have only one path, which limits available options.

xor-xor: Done.

szok (approved):

PR #570, opened 2016-12-01
Sort anomalies by date, usage type (+ 2 small bugfixes).

xor-xor:


src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -339,6 +340,45 @@ def group_missing_values_by_date(usage_type, missing_values):
...
+    for owner in anomalies.keys():
+        ret[owner]['missing_values'] = anomalies[owner]['missing_values']
+        changes = anomalies[owner]['unusual_changes']
+        changes_ = []
+        for ut, ch in changes.items():
+            for ch_ in ch:
+                t = (ch_[0], ch_[1], ut, ch_[2], ch_[3], ch_[4])
+                changes_.append(t)
+        # Key used here will be like: "2016-10-06:2016-10-07:Sample UsageType"
+        changes_.sort(key=lambda c: "{}:{}:{}".format(c[0], c[1], c[2]))
```
mkurek: return tuple `(c[0], c[2])` (there is no need to sort by end date - it's always start date + 1 day, isn't it?

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -339,6 +340,45 @@ def group_missing_values_by_date(usage_type, missing_values):
...
+    `missing_values` subdict remains unmodified.
+    """
+    ret = defaultdict(lambda: {'missing_values': {}, 'unusual_changes': {}})
+    for owner in anomalies.keys():
+        ret[owner]['missing_values'] = anomalies[owner]['missing_values']
+        changes = anomalies[owner]['unusual_changes']
+        changes_ = []
+        for ut, ch in changes.items():
+            for ch_ in ch:
+                t = (ch_[0], ch_[1], ut, ch_[2], ch_[3], ch_[4])
```
mkurek, line 376: could we make it more 'developer-friendly'? What are these indexes? Are these all of them (or we're skipping some value)?

xor-xor: Well, we iterate here over `changes`, so the meaning of `ch` should be obvious :wink:

And if you'd look into the docstring here, you'd notice that we basically take one such `ch` (which is a tuple of 5 elements), and insert usage type (`ut`) in the middle of them - and since this textual description is just 8 lines apart from all this, I've just assumed that'd suffice.

I can add some comment above this line saying that we are basically inserting usage type in the middle of `ch` tuple, though - would that be enough for you?

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: That's true, `c[1]` is not needed here - I was going to be explicit here, that's all.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Well, we iterate here over `changes`, so the meaning of `ch` should be obvious :wink:

And if you'd look into the docstring here, you'd notice that we basically take one such `ch` (which is a tuple of 5 elements), and insert usage type (`ut`) in the middle of them - and since this textual description is just 8 lines apart from all this, I've just assumed that'd suffice.

I can add some comment above this line saying that we are basically inserting usage type in the middle of `ch` tuple, though - would that be enough for you?

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
mkurek: ok

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
mkurek: ok

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: I mean, it should be `(c[0], c[1])` :wink: (i.e. `c[2]` *not* `c[1]` is not needed here)

xor-xor: Done.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Done.

mkurek (approved):

szok (approved):

PR #576, opened 2016-12-09
[API] Add endpoints for listing/fetching PricingServices.

xor-xor:


src/ralph_scrooge/media/api_schema.yaml
```diff
@@ -647,23 +673,19 @@ paths:
         - v0.10
   /v0.10/usage-types/:
     get:
-      summary: Fetch all usage types
+      description: "This endpoint doesn't require authentication."
+      summary: List all usage types
```
ghost, line 677: `description` has double quote but `summary` doesn't?

src/ralph_scrooge/media/api_schema.yaml
```diff
@@ -488,7 +515,7 @@ paths:
           description: Created
       tags:
         - v0.9
-  '/v0.9/pricingserviceusages/{pricing_service_id}/{usages_date}/':
+  /v0.9/pricingserviceusages/{pricing_service_id}/{usages_date}/:
```
ghost, line 518: literally "date", I mean not range or something?

src/ralph_scrooge/media/api_schema.yaml
xor-xor: Yep. Both cases are interpreted as strings in YAML anyway, so this double quote is not really needed in case of `description`, but since it already contains `'` character, I've added them to avoid confusion for editors, who will be looking for closing/matching `'`.

src/ralph_scrooge/media/api_schema.yaml
xor-xor: Yep, that endpoint returns usages for a given day (e.g. `2016-12-09`), not range.

ghost (approved):

szok (approved):

PR #577, opened 2016-12-09
Add auto-insertion of 'host' param to API schema (+ small cleanup).

xor-xor:


src/ralph_scrooge/management/commands/generate_json_schema.py
```diff
@@ -47,6 +47,14 @@ def handle(self, *args, **options):
                     "Aborting.".format(e)
                 )
                 return
+        if settings.SCROOGE_HOST:
```
szok, line 50: There will be better?
```
if not settings.SCROOGE_HOST:
    logger....
    return

schema['host'] = settings.SCROOGE_HOST
```

src/ralph_scrooge/settings/base.py
```diff
@@ -366,8 +366,10 @@
 EMAIL_NOTIFICATIONS_SENDER = ''
 EMAIL_USE_TLS = False
 
-# Base URL component for constructing Scrooge's links in outgoing mails.
-BASE_MAIL_URL = ''
+# E.g. 'scrooge.mydomain.com', 'localhost:8000' - this setting will be used for
+# cases when you need to have an access to Scrooge's hostname, but you don't
+# have an acces to HTTP request or socket.
+SCROOGE_HOST = ''
```
ghost, line 374: I'd add note if this is required or optional

ghost (approved):

src/ralph_scrooge/management/commands/generate_json_schema.py
xor-xor: That wouldn't handle the case where you'd have `host` specified directly/manually in schema file (which may be much more convenient during development).

src/ralph_scrooge/settings/base.py
xor-xor: Done.

szok (approved):

PR #584, opened 2016-12-21
Fix incorrect behaviour of left menu

ar4s:
Revert to prevoious version of ``leftmenu.html``


szok (approved):

quamilek (commented):
What does it mean? What this behavior?

xor-xor (approved):

PR #588, opened 2017-02-17
Support plugin

mkurek:
Sync supports from Ralph3.

### DIFF
--- a/src/ralph_scrooge/plugins/collect/ralph3_support.py
+++ b/src/ralph_scrooge/plugins/collect/ralph3_support.py
@@ -22,11 +22,17 @@ def _get_object_id_from_url(url):
 
 
 def update_support(support):
+    """
+    For each base object assigned to support, create (or update or delete)
+    `SupportCost` with info about support start and end and price (price is
+    divided equally between all objects attached to the support).
+    """
     logger.info('Processing support {}'.format(support['__str__']))
     price = float(support['price'])
     objects_ids = map(_get_object_id_from_url, support['base_objects'])
     support_cost_ids = set()
     if len(objects_ids) > 0:
+        # divide cost equally for each object
         price_per_object = price / len(objects_ids)
         for obj_id in objects_ids:
             try:
@@ -52,6 +58,7 @@ def update_support(support):
                     )
                 )
                 support_cost_ids.add(support_cost.pk)
+    # delete SupportCost for objects which no longer exists
     supports_costs_to_delete = SupportCost.objects.filter(
         support_id=support['id']
     ).exclude(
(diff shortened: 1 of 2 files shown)


szok (approved):

src/ralph_scrooge/tests/plugins/collect/test_ralph3_support.py
```diff
@@ -0,0 +1,114 @@
...
+def _get_pricing_object_url(obj_id):
+    return 'http://ralph.allegro/api/base-objects/{}'.format(obj_id)
+
+
+class TestRalph3SupportPlugin(ScroogeTestCase):
+    def setUp(self):
+        self.today = date(2017, 2, 17)
+        self.daily_asset1 = DailyAssetInfoFactory()
+        self.daily_asset2 = DailyAssetInfoFactory()
+        self.daily_asset3 = DailyAssetInfoFactory.build()
```
xor-xor: Why this one has `.build()` while two others don't..? (just curious)

src/ralph_scrooge/tests/plugins/collect/test_ralph3_support.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assertEquals(support.end, self.data['date_to'])
+        self.assertEquals(support.remarks, self.data['__str__'])
+        self.assertEquals(support.pricing_object.id, pricing_object.id)
+        cost = float(self.data['price']) / (
+            divide_to or len(self.data['base_objects'])
+        )
+        self.assertEquals(support.cost, cost)
+        self.assertEquals(support.forecast_cost, cost)
+
+    def test_update_support(self):
```
xor-xor: What about using some more descriptive names for test cases..? (i.e. what do you expect..? in what circumstances..? etc.)

src/ralph_scrooge/tests/plugins/collect/test_ralph3_support.py
mkurek: I've moved it to the function where this is used (it's used only in one function). And the reason is: in this case, we want to create object but without saving it to DB to test object from ralph which is not present in Scrooge.

src/ralph_scrooge/plugins/collect/ralph3_support.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def _get_object_id_from_url(url):
+    """
+    Extract object ID from Ralph API url.
+    """
+    return url.rstrip('/').rpartition('/')[2]
+
+
+def update_support(support):
```
xor-xor, line 24: Consider adding some docstring here briefly describing what this fn does (e.g. what does it mean "update" in this context? what gets changed in `support`?).

xor-xor (approved):

PR #591, opened 2017-02-21
Cycle detection

mkurek:
Implementation of PricingService charing cycle detection algorithm.
Steps:
* build graph of dependency between PricingServices for given date
* (try to) run DFS from each PricingService looking for cycle

In next PRs I'll expose it in GUI as well as add some tests for cycles detection

### DIFF
--- a/src/ralph_scrooge/management/commands/detect_cycles.py
+++ b/src/ralph_scrooge/management/commands/detect_cycles.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+from datetime import datetime
+
+from django.core.management.base import BaseCommand
+
+from ralph_scrooge.utils.cycle_detector import detect_cycles
+
+
+def valid_date(date_):
+    try:
+        return datetime.strptime(date_, "%Y-%m-%d").date()
+    except ValueError:
+        raise argparse.ArgumentTypeError(
+            "Invalid date: '{}'.".format(date_)
+        )
+
+
+class Command(BaseCommand):
+    """Calculate daily costs for a given day (defaults to yesterday) and
+    pricing service (defaults to all which are bot active and have fixed
+    price).
+    """
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--date',
+            type=valid_date,
+            dest='date',
+            required=True,
+        )
+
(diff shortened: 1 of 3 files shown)


src/ralph_scrooge/management/commands/detect_cycles.py
```diff
@@ -0,0 +1,45 @@
...
+    try:
+        return datetime.strptime(date_, "%Y-%m-%d").date()
+    except ValueError:
+        raise argparse.ArgumentTypeError(
+            "Invalid date: '{}'.".format(date_)
+        )
+
+
+class Command(BaseCommand):
+    """Calculate daily costs for a given day (defaults to yesterday) and
```
xor-xor: Are you sure that this description is relevant to what is happening here..? (i.e. calculation of costs != cycle detection, but maybe I'm missing something..?)

src/ralph_scrooge/utils/cycle_detector.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Perform DFS on PricingServices dependecy graph looking for cycles
+
+    Params:
+        node: PricingService
+        graph: dict with adjacency list of PricingServices
+        visited: set of already visited PricingServices during graph traversal
+        ps_stack: stack of PricingServices in current traversal (top->down)
+    Returns:
+        list of lists of PricingServices
```
xor-xor: Add a note what is returned when there's no cycle.

src/ralph_scrooge/utils/cycle_detector.py
```diff
@@ -0,0 +1,65 @@
...
+    graph = defaultdict(list)
+    for ps in PricingService.objects.all():
+        if ps.plugin_type == PricingServicePlugin.pricing_service_fixed_price_plugin:  # noqa: E501
+            continue
+        for dep in ps.get_dependent_services(date):
+            graph[dep].append(ps)
+    return dict(graph)
+
+
+def _detect_cycles(node, graph, visited, ps_stack):
```
xor-xor, line 20: What about writing some basic tests for this function..?

xor-xor (changes requested):

src/ralph_scrooge/utils/cycle_detector.py
xor-xor: s/list of lists of PricingServices/list of lists of PricingServices (or empty list if there's no cycle)/

xor-xor (approved):

PR #592, opened 2017-02-22
[API] service-environments-costs: Return total_costs as null when costs are not accepted

mkurek:
Return null instead of 0 for days (months), where costs were not accepted

### DIFF
--- a/src/ralph_scrooge/rest_api/public/v0_10/service_environment_costs.py
+++ b/src/ralph_scrooge/rest_api/public/v0_10/service_environment_costs.py
@@ -385,7 +385,9 @@ def fetch_costs(
             # costs for particular date were accepted, but this service-env
             # does not have any costs then
             # TODO(mkurek): consider accepting only part of the month
-            # (currently only first day of month is checked)
+            # (currently only first day of month is checked when grouping by
+            # month - see if above - only first day of each month is placed in
+            # date_range_ )
             0 if date_ in filtered_dates else None
         )
         costs_for_date = {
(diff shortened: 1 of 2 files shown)


src/ralph_scrooge/rest_api/public/v0_10/service_environment_costs.py
```diff
@@ -378,7 +378,16 @@ def fetch_costs(
...
     for date_ in date_range_:
-        total_cost_for_date = total_costs.get(date_, 0)
+        total_cost_for_date = total_costs.get(
+            date_,
+            # None, when costs for particular date were not accepted (if
+            # accepted costs were requested) otherwise 0, which means, that
+            # costs for particular date were accepted, but this service-env
+            # does not have any costs then
+            # TODO(mkurek): consider accepting only part of the month
+            # (currently only first day of month is checked)
```
xor-xor: Could you expand this last comment a little bit by e.g. providing a link to the part of the code that "only first day of the month is checked") refers to..?

src/ralph_scrooge/tests/rest_api/public/test_service_environment_costs.py
```diff
@@ -946,3 +951,141 @@ def test_if_all_costs_are_returned_when_accept_only_param_set_to_false_given(sel
...
+        )
+        # 2016-12-14 is accepted, but sice accepted_only is set to False,
+        # total_cost will be NOT None
+        self.assertIsNotNone(
+            costs['service_environment_costs'][1]['total_cost']
+        )
+
+    @data(False, True)  # `forecast` param
+    def test_if_only_accepted_costs_are_returned_when_accept_only_param_given_and_grouped_by_month(self, forecast):  # noqa: E501
+        months = [
```
xor-xor, line 966: Could you briefly describe what is the final result of this 966-1000 block of code..? (the same applies for that other test here)

xor-xor (approved):
It's OK for me, only two small remarks re: comments.

mkurek:
@xor-xor take a look now

PR #595, opened 2017-03-01
Provide initial version of Vagrantfile.

xor-xor:


vagrant/provision.sh
```diff
@@ -0,0 +1,55 @@
...
+sudo -H -u vagrant sh -c "NVM_DIR=/home/vagrant/.nvm \
+                          . /home/vagrant/.nvm/nvm.sh && \
+                          nvm install node && \
+                          cd /home/vagrant/sources/ralph_scrooge && \
+                          make install_ui"
+
+exec sudo -u vagrant /bin/bash - <<EOF
+set -e
+
+# TODO(xor-xor): Replace all occurences of 'ralph' with 'scrooge' here when ready.
```
mkurek: when it will be ready? as far as i know scrooge is now totally separated from ralph and could use its own db and user :)

xor-xor: Yeah, right - I'll change that.

mkurek (approved):

PR #597, opened 2017-03-07
Provide ability to acknowledge detected usage anomalies.

xor-xor:
It also adds a management command `delete_old_usage_anomaly_acks` which should be added to crontab (by default, it deletes acks older than 2 months, but this number can be changed via `-m` option).


src/ralph_scrooge/management/commands/delete_old_usage_anomaly_acks.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def handle(self, num_months, *args, **options):
+        min_date = datetime.date.today() - relativedelta(months=num_months)
+        logger.info(
+            'Deleting UsageType anomaly acknowledgements older than '
+            '{:%Y-%m-%d}...'.format(min_date)
+        )
+        qs = UsageAnomalyAck.objects.filter(anomaly_date__lt=min_date)
+        qs_count = qs.count()
+        if qs.exists():
```
mkurek: why not `if qs_count`? you don't need then extra query then (I know, I know - uber optimisation in the most important piece of scrooge 👍 )

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
```diff
@@ -248,12 +240,32 @@ def _postprocess_for_report(usage_types, missing_values, unusual_changes):
...
     anomalies__ = _group_anomalies_by_owner(anomalies_)
     anomalies___ = _sort_unusual_changes_by_date_and_type(anomalies__)
     return anomalies___
 
 
+def _filter_out_ack(unusual_changes):
+    """Filter out unusual changes that are already acknowledged by owner(s) of
+    given UsageType(s).
+    """
+    unusual_changes_ = {}
```
mkurek, line 254: consider sth like:
```
unusual_changes_ = {}
for ut in unusual_changes.keys():
    ut_changes = unusual_changes[ut]
    acks = set(UsageAnomalyAck.objects.filter(
        type=ut, anomaly_date__in=[change[0] for change in ut_changes]
    ).values_list('anomaly_date', flat=True))
    unusual_changes_[ut] = [change for change in changes if change[0] not in acks]
return unusual_changes_
```

Wouldn't it be easier (and faster)?

src/ralph_scrooge/urls.py
```diff
@@ -117,7 +118,7 @@
     url(
         r'^$',
         login_required(BootstrapAngular.as_view()),
-        name='services_costs_report',
+        name='services_costs_report',  # XXX why this URL is called like that?
```
mkurek: check if it is used somewhere and you could rename it

src/ralph_scrooge/views/anomalies_ack.py
```diff
@@ -0,0 +1,52 @@
...
+            validate_date(date)
+        except (UsageType.DoesNotExist, ArgumentTypeError):
+            return False
+        if user not in ut.owners.all():
+            return False
+        return True
+
+    # POST would be an overkill here, really...
+    def get(self, request):
+        ut_id = request.GET.get('ut_id')
```
mkurek: can we put these params into url, to call it like `http://scrooge/ack/<ut_id>/<date>`, ex. `http://scrooge/ack/10/2017-03-09`?

mkurek (changes requested):

src/ralph_scrooge/management/commands/delete_old_usage_anomaly_acks.py
xor-xor: Yep, right. Fixed.

src/ralph_scrooge/management/commands/detect_usage_anomalies.py
xor-xor: Yes, probably ;) (fixed)

src/ralph_scrooge/views/anomalies_ack.py
xor-xor: I did that at first, but then I started to think - "what's the point of structuring this URL like that, since this endpoint is private and shouldn't be 'discoverable'..?" - and I think that this question remains valid ;)

src/ralph_scrooge/urls.py
xor-xor: Done.

src/ralph_scrooge/views/anomalies_ack.py
mkurek: Being private doesn't mean that we don't have to design this url properly :) IMO it has nothing to do with discoverability - rather validation and easiness.

xor-xor: But we fully control the shape of the links in notifications emails and users are not expected to enter them manually (actually, they shouldn't do that at all), so what's the point..? I.e. validation of what and easiness for who..? ;)

xor-xor: OK, I've changed it.

mkurek (approved):

PR #599, opened 2017-03-24
Add validation of data for recalculation of monthly costs report.

xor-xor:


src/ralph_scrooge/plugins/validations.py
```diff
@@ -0,0 +1,192 @@
...
+
+        self.active_teams = Team.objects.filter(active=True)
+        self.active_usage_types = UsageType.objects.filter(active=True)
+
+    # XXX And also, maybe check extra costs and dynamic extra costs..?
+    def _check_for_required_costs_and_prices(self):
+        """There should be no active UsageType without all required
+        costs/prices."""
+        q = {'start__lte': self.date, 'end__gte': self.date}
+        for ut in self.active_usage_types:
```
mkurek: it's not so simple IMO. We should exclude here `UsageType`s with `usage_type == Service Usage Type` attached to `PricingService` with plugin type `fixed price`.

We could approach this from different side: take here all `PricingService`s with `fixed_price` plugin type and check if every `UsageType` attached to them has price defined.

src/ralph_scrooge/plugins/validations.py
```diff
@@ -0,0 +1,198 @@
...
+                    'time allocated for team "{}" does not sum up to 100% '
+                    '(it\'s {}%)'
+                    .format(team.name, sum_)
+                )
+
+    def _check_usage_types(self):
+        """There should be no active usage type without usage(s) saved for
+        a given day.
+        """
+        # XXX(mkurek): only active or maybe active *and* linked to some
```
mkurek: 👍 we should take here only:

`UsageType`s with `BU` type
or 
`UsageType`s with `SU` type AND assigned to some `PricingService`

src/ralph_scrooge/plugins/validations.py
```diff
@@ -0,0 +1,198 @@
...
+        # pricing service..?
+        for ut in self.active_usage_types:
+            if not DailyUsage.objects.filter(date=self.date, type=ut).exists():
+                self.errors.append(
+                    'no usage(s) uploaded for usage type "{}"'.format(ut.name)
+                )
+
+    def _check_usage_types_percent(self):
+        """There should be no usage type where percents do not sum up to 100%.
```
mkurek: usage type -> pricing service?

mkurek (changes requested):

src/ralph_scrooge/plugins/validations.py
xor-xor: You mean something like this:
```
for ut in self.active_usage_types.filter(usage_type__in=['BU', 'SU']):
    if ut.usage_type == 'SU':
        if not ServiceUsageTypes.objects.filter(
            usage_type=ut,
            start__lte=self.date,
            end__gte=self.date,
        ).exists():
            continue
    if not DailyUsage.objects.filter(date=self.date, type=ut).exists():
        self.errors.append(
            'no usage(s) uploaded for usage type "{}"'.format(ut.name)
        )
```
?

src/ralph_scrooge/plugins/validations.py
xor-xor: Changed to:   "There should be no pricing service with usage type(s) that together do not sum up to 100%.".

src/ralph_scrooge/plugins/validations.py
mkurek: looks good

src/ralph_scrooge/plugins/cost/collector.py
```diff
@@ -134,6 +135,7 @@ def process(
         forecast=False,
         delete_verified=False,
         plugins=None,
+        from_gui=False,
```
mkurek: could you name it more general, ex. `perform_validation`?

src/ralph_scrooge/plugins/validations.py
```diff
@@ -0,0 +1,213 @@
...
+    UsageType,
+    Warehouse,
+)
+
+
+class DataForReportValidationError(Exception):
+    """`errors` should be a list containing error msgs given by Validator."""
+
+    def __init__(self, message, errors):
+        setattr(self, 'errors', errors)
```
mkurek: do you have to use `setattr` here? wouldn't `self.errors = errors` work?

src/ralph_scrooge/plugins/validations.py
```diff
@@ -0,0 +1,213 @@
...
+        Exception.__init__(self, message)
```
mkurek: why not `super`?

src/ralph_scrooge/plugins/validations.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def __init__(self, date, forecast=False):
+        self.date = date
+        self.forecast = forecast
+        self.errors = []
+
+        self.active_teams = Team.objects.filter(active=True)
+        self.active_usage_types = UsageType.objects.filter(active=True)
+
+    # XXX(mkurek): And also, what about checking (dynamic) extra costs..?
```
mkurek: 👍 - there should be two checks for dynamic extra costs:
1) if cost is defined
2) if usage type (using which cost is allocated) has some daily usages for particular day

So it should be combination of validator for extra cost and usage type

src/ralph_scrooge/plugins/validations.py
```diff
@@ -0,0 +1,213 @@
...
+            plugin_type=PricingServicePlugin.pricing_service_fixed_price_plugin,  # noqa: E501
+        ):
+            for ut in ps.usage_types.all():
+                q.update({'type': ut})
+                field = 'cost' if not self.forecast else 'forecast_cost'
+                sum_cost = UsagePrice.objects.filter(**q).aggregate(
+                    s=Sum(field)
+                )['s']
+                field = 'price' if not self.forecast else 'forecast_price'
+                sum_price = UsagePrice.objects.filter(**q).aggregate(
```
mkurek: you could combine this into single query (end of [this](https://docs.djangoproject.com/en/1.10/topics/db/aggregation/#generating-aggregates-over-a-queryset) paragraph ):
```
cost_field = 'cost' if not self.forecast else 'forecast_cost'
price_field = 'price' if not self.forecast else 'forecast_price'
up = UsagePrice.objects.filter(**q).aggregate(sum_cost=Sum(cost_field), sum_price=Sum(price_field))
sum_cost, sum_price = up['sum_cost'], up['sum_price']
```

src/ralph_scrooge/plugins/validations.py
```diff
@@ -0,0 +1,213 @@
...
+                        ut.name,
+                    )
+                )
+
+    def _check_team_costs(self):
+        """There should be no active team with undefined/zero costs on given
+        day.
+        """
+        if self.forecast:
+            q = 'forecast_cost'
```
mkurek: q -> cost_field

src/ralph_scrooge/rest_api/private/monthly_costs.py
```diff
@@ -256,20 +272,26 @@ class DailyCostsJob(WorkerJob):
     _return_job_meta = True
 
     @classmethod
-    def run(cls, day, forecast):
+    def run(cls, day, forecast, from_gui=False):
```
mkurek: imo we don't need this param here

src/ralph_scrooge/rest_api/private/monthly_costs.py
```diff
@@ -256,20 +272,26 @@ class DailyCostsJob(WorkerJob):
...
         """
         Run collecting costs for one day.
         """
         collector = Collector()
         result = {}
+        validation_errors = []
         try:
-            result = collector.process(day, forecast)
+            result = collector.process(day, forecast, from_gui=from_gui)
```
mkurek: imo `from_gui` could be always `True` here

src/ralph_scrooge/tests/plugins/test_validations.py
```diff
@@ -0,0 +1,250 @@
...
+        CostDateStatusFactory(date=self.day, accepted=True)
+
+        self.validator._check_for_accepted_costs()
+        self.assertEqual(len(self.validator.errors), 1)
+        self.assertIn('costs already accepted', self.validator.errors)
+
+    # We don't test `_check_for_cycles` method, because it's only a wrapper on
+    # `detect_cycles`, which has its own suite of tests.
+
+    # XXX(mkurek): Do we need/want to test all above with `forecast=True`
```
mkurek: probably not - but if you like to you could use subclass and set `self.forecast = True/False` in subclasses - then use `self.forecast` in proper tests :)

src/ralph_scrooge/rest_api/private/monthly_costs.py
```diff
@@ -107,16 +110,23 @@ def get(self, request, job_id, *args, **kwargs):
         progress, data, job, meta = self.run_on_worker(
             **job.kwargs
         )
+        validation_errors = meta['validation_errors']
         status = 'running'
         data = data or {}
-        data = sorted(data.items(), key=lambda k: k[0]),
-        data = [(str(i[0].date()), i[1]) for i in data[0]]
+        data_ = []
+        # XXX(mkurek): What does True/False in data mean?
```
mkurek: data is True/False and means, that job ended successfully or not

mkurek (changes requested):

src/ralph_scrooge/plugins/cost/collector.py
xor-xor: Done.

src/ralph_scrooge/plugins/validations.py
xor-xor: What's the difference, effectively..? ;) (but OK, I'll change that)

src/ralph_scrooge/plugins/validations.py
xor-xor: Done.

src/ralph_scrooge/plugins/validations.py
xor-xor: Right, thx for suggestion.

src/ralph_scrooge/plugins/validations.py
xor-xor: Done.

src/ralph_scrooge/plugins/validations.py
mkurek: just cleanliness of the code :)

src/ralph_scrooge/rest_api/private/monthly_costs.py
xor-xor: OK, thx.

src/ralph_scrooge/rest_api/private/monthly_costs.py
xor-xor: Yeah, right - I didn't know that this method is used only by GUI.

src/ralph_scrooge/rest_api/private/monthly_costs.py
xor-xor: Removed.

src/ralph_scrooge/tests/plugins/test_validations.py
xor-xor: Good idea (done).

src/ralph_scrooge/plugins/validations.py
xor-xor: Done.

src/ralph_scrooge/plugins/cost/collector.py
```diff
@@ -150,6 +154,9 @@ def process(
             date,
         ))
         self._verify_accepted_costs(date, forecast, delete_verified)
+        if perform_validation:
```
mkurek: could you add some switch to settings for it too (to easily turn it off in case of any undesired behaviour, especially on first recalculation). It could be set to `True` by default.

xor-xor: Good idea (done).

mkurek (approved):

PR #600, opened 2017-04-03
Recalculate daily costs after each usages upload.

xor-xor:


src/ralph_scrooge/rest_api/public/v0_9/pricing_service_usages.py
```diff
@@ -405,12 +408,18 @@ def get_usages(usages_date, pricing_service, filter_by_service=None):
...
-        save_usages(deserializer.validated_data)
+        _save_usages_and_recalculate_costs(deserializer.validated_data)
         return HttpResponse(status=201)
     return Response(deserializer.errors, status=400)
 
 
 @transaction.atomic
+def _save_usages_and_recalculate_costs(pricing_service_usage):
+    save_usages(pricing_service_usage)
+    # TODO(xor-xor): Consider recalculation of forecast costs as well.
```
mkurek: 👍

src/ralph_scrooge/rest_api/public/v0_9/pricing_service_usages.py
```diff
@@ -405,12 +408,18 @@ def get_usages(usages_date, pricing_service, filter_by_service=None):
...
+    _recalculate_costs(pricing_service_usage['date'])
```
mkurek: this should be executed only if usages are saved for PricingService with fixed prices - otherwise we could get "ddosed" by pushing to API changes for other Pricing Services, especially the ones with `overwrite=False`.

mkurek (changes requested):

src/ralph_scrooge/rest_api/public/v0_9/pricing_service_usages.py
xor-xor: Done.

src/ralph_scrooge/rest_api/public/v0_9/pricing_service_usages.py
xor-xor: Done.

mkurek (approved):

PR #612, opened 2017-04-28
Admin improvement (most important: autocomplete)

mkurek:
Several improvements to Scrooge admin interface:
* autocomplete for (almost) all foreign keys (including inlines) - some were ommited since they by-nature have low number (most tens) of items, ex. environment or warehouse
* improved listings (additional columns, filters)
* additional select_related and prefetch_related to speed-up loading of admin pages
* added ordering for some models


src/ralph_scrooge/utils/common.py
```diff
@@ -134,3 +135,14 @@ def validate_date(date_):
         raise argparse.ArgumentTypeError(
             "Invalid date: '{}'.".format(date_)
         )
+
+
+def camel_case_to_dash(name):
+    """
+    Converts CamelCase to dash-case, for example:
```
xor-xor: ...or kebab-case maybe..? :wink: 

http://stackoverflow.com/a/12273101/5768173

src/ralph_scrooge/views/autocomplete.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class ScroogeAutocomplete(autocomplete.Select2QuerySetView):
+    search_fields = None
+
+    def get_queryset(self):
+        queryset = self.model._default_manager.all()
+        if self.q:
+            # split query string by whitespaces and apply filtering using
+            # the folowing condition: each token from query string has to
```
xor-xor: Could you provide an example what would happen when this condition is not met..?

src/ralph_scrooge/utils/common.py
mkurek: as you wish :)

src/ralph_scrooge/views/autocomplete.py
mkurek: `conditon` was wrong word here - i've changed it to `logic`

xor-xor (approved):

PR #621, opened 2017-10-24
Add ``support_team`` field to usage type

ar4s:


src/ralph_scrooge/models/usage.py
```diff
@@ -129,6 +129,12 @@ class UsageType(BaseUsage):
             ' recalculating costs'
         ),
     )
+    support_team = db.TextField(
```
mkurek: why not `CharField`?

mkurek (approved):

PR #625, opened 2018-02-01
identify environment by name instead of ralph_id

MarekBleschke:


src/ralph_scrooge/models/service.py
```diff
@@ -83,19 +83,13 @@ class Meta:
...
 
-class Environment(Named.NonUnique):
+class Environment(Named):
     ci_id = db.IntegerField(
         unique=True,
         null=True,
         blank=True,
         verbose_name=_("id from cmdb"),
     )
-    ralph3_id = db.IntegerField(
```
mkurek: why are you removing this field (and, for example, leaving ci_id)?

MarekBleschke: I want to move from identifying by ralph_id to identifying environments by name. Change with integration with hermes introduced a problem because we don't have ralph_id in incoming data so it's not added to new env entry. Environment update from ralph3_service_environment.py added duplicated environment entry (as it's looking existing env by ralph_id) which caused problems.

If you prefer I can leave this field, but I thought it could be confusing.

mkurek: it's good idea to identify env by name, but I'd remove `ci_id` and `ci_uid` fields too

MarekBleschke: ok

MarekBleschke: done

src/ralph_scrooge/admin.py
```diff
@@ -336,7 +336,6 @@ class BusinessLineAdmin(UpdateReadonlyMixin, admin.ModelAdmin):
 class EnvironmentAdmin(UpdateReadonlyMixin, admin.ModelAdmin):
```
ar4s: ``UpdateReadonlyMixin`` is not necessary now

ar4s (changes requested):

src/ralph_scrooge/admin.py
MarekBleschke: removed

mkurek (approved):

ar4s (approved):

PR #638, opened 2018-07-26
Speed up service-environment sync from ralph

mkurek:


src/ralph_scrooge/utils/common.py
```diff
@@ -146,3 +153,32 @@ def camel_case_to_kebab_case(name):
     """
     s1 = re.sub('(.)([A-Z][a-z]+)', r'\1-\2', name)
     return re.sub('([a-z0-9])([A-Z])', r'\1-\2', s1).lower()
+
+
+class timeit(object):
```
ar4s, line 158: where do you use this class?

mkurek: currently nowhere, but it's really helpful in tracking down bottlenecks (both time and db queries count) in plugins

src/ralph_scrooge/utils/common.py
```diff
@@ -6,14 +6,21 @@
...
+import logging
 import re
+import time
 from datetime import datetime
 from decimal import Decimal
 
+from functools import wraps
+
 from django.conf import settings
 # DEPRECATED. Import memoize directly from `ralph_scrooge.cache`
```
ar4s: This comment is related to 20 line not to 19. It is necessary?

mkurek: moved one line below; yes - it's still relevant - there are a lot of imports from `from ralph_scrooge.utils.common import memoize`

ar4s (approved):

MarekBleschke (approved):

PR #641, opened 2018-07-26
Skip rows with only-0-values in reports

mkurek:


MarekBleschke (dismissed):

ar4s (approved):

ar4s (approved):

src/ralph_scrooge/settings/base.py
```diff
@@ -342,14 +342,18 @@
         },
     },
     'loggers': {
+        '': {
```
ar4s: #639 contains this change

ar4s (changes requested):

src/ralph_scrooge/settings/base.py
mkurek: fixed

ar4s (approved):

MarekBleschke (approved):

PR #648, opened 2018-08-21
ignore zero value usages

szymi-:

PR #652, opened 2018-12-04
Fix distributing pricing service cost when there is lack of some daily usages

mkurek:
When pricing service has multiple services (for distributing cost) and
not every DailyUsage is present (see #648 - ignoring zero-value usage)
costs were not distributed properly. It was caused by storying usage
values for single service environment on lists - these lists have
different size and when accessing by index (and using `zip` to
join this list with total usage and percentage for usage type)
usage value was combined with wrong total usage and percentage.

### DIFF
--- a/src/ralph_scrooge/plugins/cost/pricing_service.py
+++ b/src/ralph_scrooge/plugins/cost/pricing_service.py
@@ -237,12 +237,11 @@ class PricingServiceBasePlugin(BaseCostPlugin):
                 ...
             }
         """
-        usages = defaultdict(list)
-        total_usages = []
-        percentage = []
+        usages = defaultdict(dict)
+        total_usages = {}
+        percentage = {}
         result = defaultdict(list)
         self.pricing_service = pricing_service
-
         for service_usage_type in service_usage_types:
             service_excluded = excluded_services.union(
                 service_usage_type.usage_type.excluded_services.all()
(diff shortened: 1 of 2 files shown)


src/ralph_scrooge/plugins/cost/pricing_service.py
```diff
@@ -255,18 +254,22 @@ def _distribute_costs(
...
                 date=date,
                 excluded_services=service_excluded,
-            ))
-            percentage.append(service_usage_type.percent)
+            )
+            percentage[usage_type_id] = service_usage_type.percent
         # create hierarchy basing on usages
         for (po, se), po_usages in usages.items():
-            po_usages_info = zip(po_usages, total_usages, percentage)
+            po_usages_info = [
```
vi4m: what does po mean?

src/ralph_scrooge/plugins/cost/pricing_service.py
```diff
@@ -255,18 +254,22 @@ def _distribute_costs(
...
+                (value, total_usages[ut_id], percentage[ut_id])
+                for ut_id, value in po_usages.items()
```
vi4m: what does ut mean? usage_type?

src/ralph_scrooge/plugins/cost/pricing_service.py
MarekBleschke: +1, IMO using full names for `po`, `ut`, `se` would be more readable for future us ;)

mkurek: sure - i've expanded all abbreviations here

src/ralph_scrooge/plugins/cost/pricing_service.py
mkurek: yes - it's explained now

szymi- (approved):

MarekBleschke (approved):

